PR #90, opened 2019-04-17
add reload_param_defaults function

pmldrmota:


dnadlinger:
Thanks for the PR – as discussed earlier, this would have overwritten parameters that had overrides set. I've committed an alternative as https://github.com/OxfordIonTrapGroup/ndscan/commit/38a23c48bbc2656bb506ba2675b63c8510be261d. Also check out https://github.com/OxfordIonTrapGroup/ndscan/commit/81b4751a12c169209da35f103140012f53f16d5f for how to add unit tests for this sort of stuff (which I had forgotten to add to the previous commit before pushing…).

PR #93, opened 2019-04-18
enable gaussian and v-shape fit options

pmldrmota:


pathfinder49:
@pmldrmota Can this be closed?

dnadlinger:
Added in #193 including default annotations.

PR #112, opened 2019-08-12
Add oitg sinusoid fit for default analysis

AUTProgram:
I have added sinusoid to FIT_OBJECTS and updated the dictionary for default analysis


dnadlinger:
Merged as https://github.com/OxfordIonTrapGroup/ndscan/pull/115 – I ended up having to reformat a bunch of things due to upgrading YAPF in the container too.

Issue #211, opened 2020-10-23
Default analyses not executed for subscans


AUTProgram:
When running a subscan through a higher-level fragment, the default analyses are not executed as expected. Consider the following code:

```python
from artiq.experiment import EnumerationValue

from ndscan.experiment import (Annotation, CustomAnalysis, ExpFragment,
    FloatChannel, FloatParam, LinearGenerator, make_fragment_scan_exp,
    setattr_subscan)


class BaseFragment(ExpFragment):
    def build_fragment(self):
        self.setattr_param(
            "param1",
            FloatParam,
            "Parameter 1",
            default=1.0)
        self.setattr_param(
            "param2",
            FloatParam,
            "Parameter 2",
            default=3.0)
        self.setattr_param(
            "param3",
            FloatParam,
            "Parameter 3",
            default=4.0)
        self.setattr_param(
            "param4",
            FloatParam,
            "Parameter 4",
            default=8.0)

        self.setattr_result(
            "res",
            FloatChannel)

    def run_once(self):
        self.res.push(self.param1.get()*10 + self.param2.get()*20
                      + self.param3.get()*60 + self.param4.get()*80)

    def get_default_analyses(self):
        fits = [CustomAnalysis([self.param1],
                               self._create_analyse_fn(self.param1),
                               [FloatChannel("x0_fit")]),
                CustomAnalysis([self.param2],
                               self._create_analyse_fn(self.param2),
                               [FloatChannel("x0_fit")]),
                CustomAnalysis([self.param3],
                               self._create_analyse_fn(self.param3),
                               [FloatChannel("x0_fit")]),
                CustomAnalysis([self.param4],
                               self._create_analyse_fn(self.param4),
                               [FloatChannel("x0_fit")])]
        return fits

    def _create_analyse_fn(self, scanned_param):
        def _analyse_scan(axis_values, result_values, analysis_results):
            x = axis_values[scanned_param]
            y = result_values[self.res]

            analysis_results["x0_fit"].push(-1.0)

            annotations = [
                Annotation("location",
                           coordinates={
                               scanned_param: analysis_results["x0_fit"]})]

            return annotations
        return _analyse_scan


ScanBaseFragment = make_fragment_scan_exp(BaseFragment)


class SubscanFragment(ExpFragment):
    def build_fragment(self):
        self.setattr_fragment("scanned_fragment", BaseFragment)

        self.arg = self.get_argument(
            "Argument of scan",
[...]

PR #217, opened 2020-12-01
Change handling of default_analyses for subscans

AUTProgram:
Attempts to fix the problems mentioned in #211 . This is a minimal working solution that is currently in use in our lab and does not seem to have any unwanted side-effects so far.
Have run the unit tests included in ndscan, no errors there.


dnadlinger:
I realised only after fixing the CI issues (which were unrelated to your PR; rather, connected to the Python/Conda upgrades) that the implementation still had an issue: In the case of multiple possible, runtime-selected 1D scans, where each of them has an associated analysis, multiple result channels would have been created (and only one actually pushed to in `run`). In #219, I've implemented a slightly different fix which avoids creating the invalid result channels, but still returns the results from `run`.

PR #248, opened 2021-07-16
Fix issue 247: AttributeError in xy_1d.py

### LINKED ISSUE #247: xy_1d error: 'list' object has no attribute 'size'
Using a results channel with error bars I get the following backtrace generated by the applet

```
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/ndscan/plots/xy_1d.py", line 201, in _update_points
print:    s.update(x_data, points)
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/ndscan/plots/xy_1d.py", line 76, in update
print:    self.view_box.addItem(self.error_bar_item)
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 409, in addItem
print:    self.updateAutoRange()
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 890, in updateAutoRange
print:    childRange = self.childrenBounds(frac=fractionVisible)
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line 1402, in childrenBounds
print:    bounds = self.mapFromItemToView(item, item.boundingRect()).boundingRect()
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/graphicsItems/ErrorBarItem.py", line 156, in boundingRect
print:    self.drawPath()
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/graphicsItems/ErrorBarItem.py", line 86, in drawPath
print:    xs = fn.interweaveArrays(x, x)
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/functions.py", line 694, in interweaveArrays
print:    size = sum(x.size for x in args)
print:  File "/miniconda/envs/devenv/lib/python3.8/site-packages/pyqtgraph/functions.py", line 694, in <genexpr>
print:    size = sum(x.size for x in args)
print:AttributeError: 'list' object has no attribute 'size'
```

I have done a small amount of digging. It looks like the pyqtgraph `setData` method explicitly requires a np array (https://github.com/pyqtgraph/pyqtgraph/blob/b9e079b10cc95ec7344c2da95b5ac6827fb02cc5/pyqtgraph/graphicsItems/ErrorBarItem.py#L37) , but some debug prints tell me we're passing in a list. 

There is an alternative dead (?) codepath here https://github.com/OxfordIonTrapGroup/ndscan/blob/9d46a98df2abafa61640b6d8085a59491b240441/ndscan/plots/xy_1d.py#L47 that converts the x / y data into an array. Doing the same for the other codepath seems to resolve this issue.

Having talked to @dnadlinger it sounds like the x / y data ndscan uses will indeed be lists. So, by not casting to arrays we are violating the contract with pyqtgraph. Best guess is that this only ever worked by accident and some modification to pyqtgraph caused it to break (I don't feel like bisecting, so haven't confirmed...)

After discussion with @dnadlinger we've agree that we should fix this by casting x / y data to arrays for both code paths.

lochsh:
PyQtGraph's ErrorBarItem's setData method requires numpy arrays, but
ndscan was passing in lists, causing an AttributeError when the size
method was called.

Looking at PyQtGraph's history, the documentation has specified this since the `setData` method was added https://github.com/pyqtgraph/pyqtgraph/blob/8b0a866ad9a0a4807c4836c577ce65ecac5fd283/pyqtgraph/graphicsItems/ErrorBarItem.py. 

Potentially no numpy array-specific methods were being called until some other change in PyQtGraph, hence this problem occurring now.

Flake8 and YAPF checked. I would have added an automated test but there weren't any for this plotting code. I verified the fix by being able to plot data with error bars that previously caused the described AttributeError to be raised
image
Do let me know if there's any more testing I can do.

Closes #247

### DIFF
--- a/ndscan/plots/xy_1d.py
+++ b/ndscan/plots/xy_1d.py
@@ -31,8 +31,9 @@ class _XYSeries(QtCore.QObject):
 
     def update(self, x_data, data):
         def channel(name):
-            return data.get("channel_" + name, [])
+            return np.array(data.get("channel_" + name, []))
 
+        x_data = np.array(x_data)
         y_data = channel(self.data_name)
         num_to_show = min(len(x_data), len(y_data))
 
@@ -44,16 +45,13 @@ class _XYSeries(QtCore.QObject):
             return
 
         if self.plot_left_to_right:
-            x_data = np.array(x_data)
             order = np.argsort(x_data[:num_to_show])
 
-            y_data = np.array(y_data)
             self.data_item.setData(x_data[order], y_data[order])
             if self.num_current_points == 0:
                 self.view_box.addItem(self.data_item)
 
             if self.error_bar_item:
-                y_err = np.array(y_err)
                 self.error_bar_item.setData(x=x_data[order],
                                             y=y_data[order],
                                             height=y_err[order])
@@ -67,7 +65,7 @@ class _XYSeries(QtCore.QObject):
             if self.error_bar_item:
                 self.error_bar_item.setData(x=x_data[:num_to_show],
                                             y=y_data[:num_to_show],
-                                            height=(2 * np.array(y_err[:num_to_show])))
+                                            height=(2 * y_err[:num_to_show]))
                 if self.num_current_points == 0:
                     self.view_box.addItem(self.error_bar_item)
 


dnadlinger:
Many thanks for the nice PR!

> Flake8 and YAPF checked. I would have added an automated test but there weren't any for this plotting code.

Yep, the lack of coverage for all the UI code isn't great – I must admit that this is in large part because I couldn't be bothered to figure out how to best integrate this with the CI tests that all run on a headless system.

lochsh:
No problem! It takes a lot more care to have automated tests for such code, and sometimes it isn't worth it! I certainly didn't point it out disapprovingly, just to make sure I hadn't missed any relevant tests :) 

Thanks again

PR #249, opened 2021-07-21
Allow non-implemented build_fragment for super() with multiple inheritance

pmldrmota:
This would allow using `super()` with multiple inheritance in `build_fragment()`. Currently, the `Fragment`'s own `build_fragment()` method raises in the following minimal example.
```python
class F(Fragment):
    def build_fragment(self):
        super().build_fragment()

class G(Fragment):
    def build_fragment(self):
        super().build_fragment()

class C(F, G):
    def build_fragment(self):
        super().build_fragment()
```

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -259,8 +259,7 @@ class Fragment(HasEnvironment):
         :param kwargs: Any extra keyword arguments that were passed to the
             ``HasEnvironment`` constructor.
         """
-        raise NotImplementedError("build_fragment() not implemented; "
-                                  "override it to add parameters/result channels.")
+        pass
 
     def setattr_fragment(self, name: str, fragment_class: Type["Fragment"], *args,
                          **kwargs) -> "Fragment":


dnadlinger:
Could you expand on what you expect to happen in the example? Which classes would actually contain implementation code? And which method do you expect to be called by the super() call in C? (As per the standrd method resolution order, Python will call F.build_fragment(), irrespective of this PR.)

pmldrmota:
In the above example, the methods that will be called when `c = C()` is instantiated, are

1. `HasEnvironment.__init__`
2. `Fragment.build`
3. `C.build_fragment`
4. `F.build_fragment`
5. `G.build_fragment`
6. `Fragment.build_fragment`

irrespective of the PR. However, the last call currently raises the `NotImplementedError`, and to my knowledge, it can't be excluded from the mro. 
For the example, at least `F` and `G`'s `build_fragment` methods would contain implementation code, and `C` optionally.

In this simple case, the exception can be worked around manually by not using `super()` at all:
```python
class F(Fragment):
    def build_fragment(self):
        # implementation

class G(Fragment):
    def build_fragment(self):
        # implementation

class C(F, G):
    def build_fragment(self):
        F.build_fragment(self)
        G.build_fragment(self)
```

dnadlinger:
> In the above example, the methods that will be called when c = C() is instantiated, are […]

This seems incorrect; where would `G.build_fragment()` be called from? `super().build_fragment()` resolves to `F.build_fragment()`.

> In this simple case, the exception can be worked around manually by not using super() at all

I don't think there is a way around manually specifying the classes if you want both to be called, with or without this PR.

---

Either way, whatever the exact problem is you are seeing, this PR won't properly fix it, while at the same time degrading the user experience. (The exception serves as a reminder that there is no reason to use a Fragment if you don't have any subfragments/parameters/…, plus guards against typos in the `build_fragment()` method name in user code.)

In general, I'd recommend avoiding multiple inheritance in favour of composition.

dnadlinger:
> In the above example, the methods that will be called when c = C() is instantiated, are […]

<del>This seems incorrect; where would `G.build_fragment()` be called from? `super().build_fragment()` resolves to `F.build_fragment()`.</del> I was looking at the wrong example; the weird super() semantics in Python do indeed cause `super().build_fragment()` in `F` to call the implementation in `G`.

Either way, this change degrades the user experience, as the exception guards against typos in the `build_fragment()` method name in user code, and also serves as a reminder that there is no reason to use a Fragment if you don't have any subfragments/parameters/…

In general, I'd strongly recommend using composition instead of multiple inheritance with fragments. I really don't want to compromise on other aspects to enable a niche use case (that imho is best avoided anyway).

> In this simple case, the exception can be worked around manually by not using super() at all:

If you absolutely feel the need to use multiple inheritance, this would be the way to go – I wouldn't even call that a workaround, just the most obvious implementation.

PR #279, opened 2022-06-08
Close core comm connection before calling scheduler pause

mbirtwell:
If we don't do this when the scan resumes we get a connection error when we
next try and connect to the core. This is similar to what's done in
`_run_scan_on_core_device`

### DIFF
--- a/ndscan/experiment/scan_runner.py
+++ b/ndscan/experiment/scan_runner.py
@@ -118,6 +118,14 @@ class ScanRunner(HasEnvironment):
                     fragment.device_cleanup()
             finally:
                 fragment.host_cleanup()
+            if hasattr(self.core, "comm") and hasattr(self.core.comm, "close"):
+                # Host-only scans don't necessarily make a connection to the
+                # core device, but we close any connection that was made here
+                # otherwise the comm can fail after the pause.
+                # We have all the checks above incase core is an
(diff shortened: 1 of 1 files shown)


dnadlinger:
Hmm, I'm not too sure about this. Clearly, a host-only scan doesn't create a core device connection by itself, so something else must cause it to be opened in your case.

I'd agree that this could be a nice addition nevertheless, as it avoids a common user pitfall. In fact, it would probably even be a good idead to fix this in ARTIQ itself, as the issue is so common.

However, wouldn't merging this as-is break cases where `core` is a sim/dummy core device (`artiq.sim.devices.Core`)? This is an explicitly supported use case so people can play around with the (to-be-expanded) examples using a `no_hardware` device db. Seems like we'll either need detect the presence of `comm.close()` before calling it, or otherwise fix this in ARTIQ itself.

dnadlinger:
As for changing the behaviour in ARTIQ, apparently this was already done at some point, but then reverted: https://github.com/m-labs/artiq/commit/b9c3edd49f66a63f8a14878358e26741fcf9786f Not sure what the context was here.

dnadlinger:
Ah, apologies, I didn't see that you pushed an update – just merged #302, which is your commit with a slight simplification.

PR #282, opened 2022-06-08
fragment.setattr_param_rebind: accept handles rather than an owner/name

### LINKED ISSUE #212: Allow passing parameter handle when overriding a parameter
The `override_param` method of `Fragment` currently accepts an initial value to override default value of the passed parameter and then returns the newly created store underlying the overridden parameter. It would be nice if that method could also accept the handle (or maybe store) of another parameter such that the value of the overridden parameter is forced to be the same as the parameter corresponding to the passed handle.
One use case for this is forcing several parameters from subfragments to have a common value. At the moment, this seems to be only possible by creating a parameter in top-level fragment, calling `override_param` on the relevant params in subfragments, keeping the returned stores as attributes and then manually setting the values of all those stores in `device_setup`. While also making sure that `device_setup` of the subfragments is only called afterwards to have the updated value (for non-overridden parameters, this is usually taken care of automatically by ndscan during a scan, I think). This is a bit roundabout and would be easier if one could just pass the top-level parameter to the `override_param` method and let ndscan take care of the rest.

Or is there currently already an easier way of doing this in ndscan?

hartytp:
Needs testing before merge. I will do that once upstream (@pmldrmota?) confirm we're to merge a breaking change like this and confirm we're happy with the general approach.

closes #212

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -365,7 +365,10 @@ class Fragment(HasEnvironment):
         new_handle = new_param.HandleType(self, name)
         setattr(self, name, new_handle)
 
-        rebinds = [original_handle] + additional_handles
+        rebinds = [original_handle]
+        if additional_handles is not None:
+            rebinds += additional_handles
+            
         self._rebound_subfragment_params[name] = []
         for handle in rebinds:
             assert handle.name in handle.owner._free_params, (
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -352,7 +352,7 @@ class Fragment(HasEnvironment):
 
         original_owner = original_handle.owner
         original_name = original_handle.name
-        original_param = original_handle._free_params[original_name]
+        original_param = original_owner._free_params[original_name]
 
         new_param = deepcopy(original_param)
         new_param.fqn = self.fqn + "." + name
@@ -362,7 +362,7 @@ class Fragment(HasEnvironment):
         new_handle = new_param.HandleType(self, name)
         setattr(self, name, new_handle)
 
-        rebinds = [original_param] + additional_params
+        rebinds = [original_handle] + additional_handles
         self._rebound_subfragment_params[name] = []
         for handle in rebinds:
             assert handle.name in handle.owner._free_params, (
@@ -370,7 +370,7 @@ class Fragment(HasEnvironment):
             "already rebound?".format(handle.name))
             handles = handle.owner._get_all_handles_for_param(handle.name)
             del handle.owner._free_params[handle.name]
-            self._rebound_subfragment_params[name] += original_handles
+            self._rebound_subfragment_params[name] += handles
 
         return new_handle
 
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -319,7 +319,8 @@ class Fragment(HasEnvironment):
                              original_handle: ParamHandle,
                              additional_handles: Optional[List[ParamHandle]]=None,
                              **kwargs) -> ParamHandle:
-        """Create a parameter that overrides the value of subfragment parameters.
+        """Create a parameter that overrides the value of one or more subfragment
+        parameters.
 
         The most common use case for this is to specialise the operation of one or more
         generic subfragments. For example, there might be a fragment ``Fluoresce`` that
@@ -334,9 +335,11 @@ class Fragment(HasEnvironment):
             Python identifier; the new parameter handle will be accessible as
             ``self.<name>``.
         :param original_handle: handle for the overridden parameter. The new parameter
-            will inherent metadata from this parameter unless overridden by kwargs.
+            will inherent all metadata from this parameter that is not overridden by
+            kwargs.
         :param additional_handles: List of additional parameter handles to rebind.
-            Metadata from these parameters will not be inherited by the new parameter.
+            Allows one parameter to drive multiple subfragment parameters. Metadata from
+            these parameters will not be inherited by the new parameter.
         :param kwargs: Any attributes to override in the parameter metadata, which
             defaults to that of the original parameter.
         :return: The newly created parameter handle.
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -350,8 +350,7 @@ class Fragment(HasEnvironment):
         assert not hasattr(self, name), "Field '{}' already exists".format(name)
         assert original_handle.name in original_handle.owner._free_params, (
             "Field '{}' is not a free parameter of original owner; "
-            "already rebound?".format(original_handle.name)
-        )
+            "already rebound?".format(original_handle.name))
         assert name not in self._rebound_subfragment_params
 
         original_owner = original_handle.owner
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -374,6 +374,9 @@ class Fragment(HasEnvironment):
             assert handle.name in handle.owner._free_params, (
                 "Field '{}' is not a free parameter of original owner; "
                 "already rebound?".format(handle.name))
+            assert isinstance(handle, new_param.HandleType), (
+                f"Field '{handle.name}' of type {type(handle)} does not match type of "
+                f"original parameter ('{type(new_handle)}'")
             handles = handle.owner._get_all_handles_for_param(handle.name)
             del handle.owner._free_params[handle.name]
             self._rebound_subfragment_params[name] += handles
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -2,7 +2,7 @@ from artiq.language import *
 from collections import OrderedDict
 from copy import deepcopy
 import logging
-from typing import Any, Dict, List, Iterable, Type, Tuple, Union
+from typing import Any, Dict, List, Iterable, Type, Tuple, Union, Optional
 
 from .default_analysis import DefaultAnalysis
 from .parameters import ParamHandle, ParamStore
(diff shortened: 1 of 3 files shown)


hartytp:
Tested using the following test case
```python
import logging


logger = logging.getLogger(__name__)


class BaseExp(nd.ExpFragment):
    def build_fragment(self):
        self.setattr_param("param", nd.FloatParam, "a parameter", 0)
        self.setattr_result("result", nd.FloatChannel)

    def run_once(self):
        self.result.push(0)

        def _analysis_fun(axis_values, result_values, analysis_results):
            logger.info(f"{'_'.join(self._fragment_path)} analysis...")
            analysis_results['fit_result'].push(axis_values[self.param][0])

        return [
            nd.OnlineFit("line", data={"x": self.param, "y": self.result}),
            nd.CustomAnalysis(
                [self.param],
                _analysis_fun,
                [nd.FloatChannel("fit_result")]
            )
        ]

class ParentExp(nd.ExpFragment):
    def build_fragment(self):
        self.base_exps = [self.setattr_fragment(f"base{i}", BaseExp) for i in range(3)]
        self.setattr_param_rebind(
            'param',
            self.base_exps[0].param,
            [exp.param for exp in self.base_exps[1:]],
            description="a parent param"
        )

    def run_once(self):
        for base_exp in self.base_exps:
            base_exp.run_once()


class SubscanExp(nd.ExpFragment):
    def build_fragment(self):
        self.setattr_fragment('parent', ParentExp)
        nd.setattr_subscan(
            self,
            "scan",
            self.parent,
            [(self.parent, "param")],
            expose_analysis_results=True,
        )

    def run_once(self):
        scan_params = nd.LinearGenerator(-1, 1, 5, True)
        _, _, results = self.scan.run([(self.parent.param, scan_params)])

ScanBaseExp = nd.make_fragment_scan_exp(BaseExp)
ScanParentExp = nd.make_fragment_scan_exp(ParentExp)
ScanScanExp = nd.make_fragment_scan_exp(SubscanExp)
```

Runs fine.

hartytp:
Still to do before merge:
- [ ] handle case where `additional_handles is None`
- [ ] fix tests / add new tests
- [ ] check documentation around `additional_handles` is clear with a fresh head

ndscan/experiment/fragment.py
```diff
@@ -316,25 +316,30 @@ def setattr_param(self, name: str, param_class: Type, description: str, *args,
 
     def setattr_param_rebind(self,
                              name: str,
-                             original_owner: "Fragment",
-                             original_name: Union[str, None] = None,
+                             original_handle: ParamHandle,
+                             additional_handles: Optional[List[ParamHandle]] = None,
```
pmldrmota, line 320: We could consider splatting (`*additional_handles`). That way, it is explicit in the function declaration which handle is used to create the new parameter, but it also represents the idea of "rebind one or more parameters" in the way the function is called, eg `setattr_param_rebind("param", param0, param1, param2)`. What do you think?

hartytp: hmmm...I don't have a strong feeling here but to me that doesn't really achieve the aim of making it 100% explicit which parameter is the template so I (weakly) prefer the signature currently in this PR.

hartytp:
hmm... @pmldrmota do you want a specific test case for this before mergin?

hartytp:
hmmm...to write a test for this properly we want to get results from the rebound fragments which essentially needs https://github.com/OxfordIonTrapGroup/ndscan/pull/283 So I'm going to hold off adding a test for this now.

@pmldrmota what are your thoughts on merging this now?

hartytp:
> It would be possible to keep the old signature with owner and name, and just add the new keyword argument for the extra parameters. Thinking more about it, passing one handle affects all handles - isn't that confusing? Specifying the fragment owning the parameter and naming the parameter seems conceptually more correct, as the identity of the parameter is distinct from the handles, and rebinding is concerned with the parameter, not the handle. The handle remains the same after all.

I don't love the idea of making `additional_handles` a kwarg as that breaks the simple rule of "all kwargs override metadata from the original parameter". What about making `additional_handles` an extra positional argument? Does this work for you?

```python
    def setattr_param_rebind(self,
                             name: str,
                             original_owner: "Fragment",
                             original_name: Optional[str] = None,
                             additional_handles: Optional[List[ParamHandle]] = None,
                             **kwargs)  -> ParamHandle:
```

Does

ndscan/experiment/fragment.py
```diff
@@ -317,59 +317,77 @@ def setattr_param(self, name: str, param_class: Type, description: str, *args,
...
+        setattr(self, name, new_handle)
+
+        owners = [(original_owner, original_name)]
+        if additional_owners is not None:
+            owners += additional_owners
+
+        self._rebound_subfragment_params[name] = []
+        for owner, *owner_name in owners:
+            [owner_name] = owner_name if owner_name != [] else [original_name]
+            owner_name = owner_name if owner_name is not None else original_name
```
hartytp, line 378: @lochsh is there an obviously cleaner way of expressing the logic around this?

ndscan/experiment/fragment.py
```diff
@@ -317,59 +317,78 @@ def setattr_param(self, name: str, param_class: Type, description: str, *args,
...
-        :param original_owner: The fragment owning the parameter to rebind.
+        :param name: The new parameter name, to be part of its FQN. Must be a valid
+            Python identifier; the new parameter handle will be accessible as
+            ``self.<name>``.
+        :param original_owner: Fragment owning the parameter to rebind. The new
+            parameter will inherent all metadata from this parameter that is not
+            overridden by kwargs.
         :param original_name: The name of the original parameter (i.e.
             ``<original_owner>.<original_name>``). If ``None``, defaults to ``name``.
+        :param additional_parameters: List of additional parameters to be overridden
```
pmldrmota, line 344: Remove trailing whitespace

ndscan/experiment/fragment.py
```diff
@@ -317,59 +317,78 @@ def setattr_param(self, name: str, param_class: Type, description: str, *args,
...
+        new_param = deepcopy(original_param)
+        new_param.fqn = self.fqn + "." + name
         for k, v in kwargs.items():
-            setattr(param, k, v)
-        self._free_params[name] = param
-        handle = param.HandleType(self, name)
-        setattr(self, name, handle)
-
-        # Deregister it from the original owner and make sure we set the store
-        # to our own later.
```
pmldrmota, line 366: Keep the comments

pmldrmota, line 366: Keep the comments

test/fixtures.py
```diff
@@ -64,6 +64,20 @@ def run_once(self):
...
 
 
+class MultipleRebindsFragment(ExpFragment):
+    def build_fragment(self):
+        self.fragments = []
+        for idx in range(10):
+            frag = self.setattr_fragment(f"add_one_{idx}", AddOneFragment)
+            self.fragments.append(frag)
+        additional_owners = [(frag, ) for frag in self.fragments[1:]]
+        self.setattr_param_rebind("value", self.fragments[0], None, additional_owners)
```
pmldrmota, line 74: Should change the default of the rebound parameter here to a different number, e.g. `1.0`, and check if **all** subfragments return `2.0` instead of `1.0`.

test/test_experiment_fragment.py
```diff
@@ -51,6 +52,12 @@ def test_recursive_rebind_override(self):
         result = run_fragment_once(rrf)[rrf.rebound_add_one.add_one.result]
         self.assertEqual(result, 3)
 
+    def test_multiple_rebinds(self):
+        mrb = self.create(MultipleRebindsFragment, [])
+        results = run_fragment_once(mrb)
+        for fragment in mrb.fragments:
+            self.assertEqual(results[fragment.result], 1)
```
pmldrmota, line 59: See comment above - if changing the rebound default to `1`, this should be `2` in all subfragments.

pmldrmota, line 59: See comment above - if changing the rebound default to `1`, this should be `2` in all subfragments.

ndscan/experiment/fragment.py
pmldrmota: What about 
```python
owner_name = original_name if not owner_name or not owner_name[0] else owner_name
```

pmldrmota (commented):
I'd like the tests to actually check that the rebinding really works for all subfragments.
I had another idea on this feature - will address in a comment, but in principle I'd agree that this could be merged.

pmldrmota:
There is maybe a cleaner way to integrate this feature if we keep `setattr_param_rebind` as-is and simply add a method

```python
def bind_param(self,
               name: str,
               original_owner: "Fragment",
               original_name: Optional[str] = None) -> None:
    """Bind a parameter of another fragment to a parameter of this fragment.

    Can only be called during :meth:`build_fragment`.

    :param name: The parameter name, accessible as ``self.<name>``.
    :param original_owner: The fragment owning the parameter to rebind.
    :param original_name: The name of the original parameter (i.e.
        ``<original_owner>.<original_name>``). If ``None``, defaults to ``name``.
    """
    assert self._building, "Can only call bind_param() during build_fragment()"

    if original_name is None:
        original_name = name
    assert hasattr(original_owner, original_name), \
        "Original owner does not have a field of name '{}'".format(original_name)
    assert original_name in original_owner._free_params, (
        "Field '{}' is not a free parameter of original owner; "
        "already rebound?".format(original_name))

    # Ensure type is compatible.
    param = self._free_params[name]
    original_param = owner._free_params[original_name]
    assert isinstance(original_param, type(param)), (
        f"Field '{original_name}' of type {type(original_param)} does not match type of "
        f"original parameter ('{type(param)}'")

    # Deregister it from the original owner and make sure we set the store
    # to our own later.
    original_handles = original_owner._get_all_handles_for_param(original_name)
    del original_owner._free_params[original_name]
    if name not in self._rebound_subfragment_params:
        self._rebound_subfragment_params[name] = []
    self._rebound_subfragment_params[name] += original_handles
```

This would allow binding one other parameter to a parameter of that fragment at a time. The workflow would look like
```python
def build_fragment(self):
    exps = [self.setattr_fragment(f'exp_{i}', BaseExp) for i in range(3)]
    self.setattr_param_rebind('param', exps[0])
    for exp in exps[1:]:
        self.bind_param('param', exp)
```

or even without rebinding:
```python
def build_fragment(self):
    exps = [self.setattr_fragment(f'exp_{i}', BaseExp) for i in range(3)]
    for exp in exps[1:]:
        exps[0].bind_param('param', exp)
```

There is little room for ambiguity about metadata inheritance because the parameter dependency tree is built up explicitly.
What do you think?

hartytp:
Thanks for the review @pmldrmota.

> There is maybe a cleaner way to integrate this feature if we keep setattr_param_rebind as-is and simply add a method

I don't have a strong opinion here. My weak opinion is that it's better to do all the rebinding in one go rather than to have separate `rebind` and a `bind` methods. I also don't feel that those method names do a great job of capturing what's going on here. But it's not something I feel strongly about.

Agreed re all the other feedback. Let's make a call on the above point and the rest is quick to implement.

hartytp:
> owner_name = original_name if not owner_name or not owner_name[0] else owner_name

Note to self: `[owner_name] = owner_name if any(owner_name) else [original_name]`

pmldrmota:
I also don't have a strong opinion there. I like the way it is now with the few small changes from the review.

hartytp:
> I also don't have a strong opinion there. I like the way it is now with the few small changes from the review.

Me too! Given the relatively minor changes remaining, I'll wait for feedback from @dnadlinger before doing anything further.

dnadlinger:
Hmm… `setattr_param_rebind` is conceptually a bit dodgy, as it combines two operations, that of creating a parameter using another one as a template, and then overriding a parameter based on another. This is not the only way of combining the two operations, although a reasonably common one (even though `setattr_param_rebind` is probably overused in our code base; a carry-over from when we didn't have `get_always_shown_params`).

The obvious solution would seem to be to divide up the two operations and expose them separately. The overriding step can then be repeated with as many different other parameters as desired (or even without using the `setattr_param_like` function, or whatever is a good name for the first part). This approach, which was more or less suggested by the initial post in #212, also sidesteps the discussion regarding which parameter to use as a template, etc.

hartytp:
> The obvious solution, which I'd prefer to piling more functionality onto setattr_param_rebind, would seem to be to divide up the two operations and expose them separately. The overriding step can then be repeated with as many different other parameters as desired (or even without using the setattr_param_like function, or whatever is a good name for the first part). This approach, which was more or less suggested by the initial post in https://github.com/OxfordIonTrapGroup/ndscan/issues/212, also sidesteps the discussion regarding which parameter to use as a template, etc.

So, does that look like:
- `setattr_param_rebind` is left unchanged and maybe becomes depreciated at some point
- `setattr_param` gets a new optional `like` positional argument which takes a parameter handle to be used as a template for the newly created parameter. This could also be a new `setattr_param_like` method but it feels cleaner to me to extend the existing method.
- `Fragment` gets a new method which looks essentially like @pmldrmota's suggestion [above](https://github.com/OxfordIonTrapGroup/ndscan/pull/282#issuecomment-1151550889)

dnadlinger:
Many thanks again for kickstarting the discussion – merged https://github.com/OxfordIonTrapGroup/ndscan/pull/300 instead, which implements the approach laid out in the last couple of comments.

hartytp:
Wonderful, thanks again @dnadlinger @pmldrmota

PR #283, opened 2022-06-08
ExpFragment.get_default_analyses: propagate analyses from subfragments

hartytp:
See https://github.com/OxfordIonTrapGroup/ndscan/issues/212 for discussion.

Needs testing before merging. Putting the PR in now so we can discuss if a feature along these lines is desirable.

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -3,8 +3,9 @@ from collections import OrderedDict
 from copy import deepcopy
 import logging
 from typing import Any, Dict, List, Iterable, Type, Tuple, Union
+import functools
 
-from .default_analysis import DefaultAnalysis
+from .default_analysis import DefaultAnalysis, CustomAnalysis
 from .parameters import ParamHandle, ParamStore
 from .result_channels import ResultChannel, FloatChannel
 from .utils import path_matches_spec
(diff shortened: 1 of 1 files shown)


hartytp:
Tested with the following test case
```python
import logging


logger = logging.getLogger(__name__)


class BaseExp(nd.ExpFragment):
    def build_fragment(self):
        self.setattr_param("param", nd.FloatParam, "a parameter", 0)
        self.setattr_result("result", nd.FloatChannel)

    def run_once(self):
        self.result.push(0)

        def _analysis_fun(axis_values, result_values, analysis_results):
            logger.info(f"{'_'.join(self._fragment_path)} analysis...")
            analysis_results['fit_result'].push(axis_values[self.param][0])

        return [
            nd.OnlineFit("line", data={"x": self.param, "y": self.result}),
            nd.CustomAnalysis(
                [self.param],
                _analysis_fun,
                [nd.FloatChannel("fit_result")]
            )
        ]

class ParentExp(nd.ExpFragment):
    def build_fragment(self):
        self.base_exps = [self.setattr_fragment(f"base{i}", BaseExp) for i in range(3)]
        self.setattr_param_rebind(
            'param',
            self.base_exps[0].param,
            [exp.param for exp in self.base_exps[1:]],
            description="a parent param"
        )

    def run_once(self):
        for base_exp in self.base_exps:
            base_exp.run_once()


class SubscanExp(nd.ExpFragment):
    def build_fragment(self):
        self.setattr_fragment('parent', ParentExp)
        nd.setattr_subscan(
            self,
            "scan",
            self.parent,
            [(self.parent, "param")],
            expose_analysis_results=True,
        )

    def run_once(self):
        scan_params = nd.LinearGenerator(-1, 1, 5, True)
        _, _, results = self.scan.run([(self.parent.param, scan_params)])

ScanBaseExp = nd.make_fragment_scan_exp(BaseExp)
ScanParentExp = nd.make_fragment_scan_exp(ParentExp)
ScanScanExp = nd.make_fragment_scan_exp(SubscanExp)
```

Without this patch no analyses are executed in `ScanParentExp`. With this patch all child analyses are executed correctly.

dnadlinger:
As discussed in #212, I'd prefer to put something like this into a separate class/… catering specifically to the use case where several `ExpFragment`s are just invoked one after another and results collected, as this is somewhat of a specific use case.

As for the implementation, if you'll allow me a bit of snark, this is why I'm sometimes pessimistic about the use of Python in large-scale software projects. :P Rather than explicitly testing for a particular type using `isinstance` and reaching into the internals, the way to go that would also work with other (incl. user-provided) `DefaultAnalysis` implementations would be to provide a `DefaultAnalysis` wrapper/proxy object that just rewrites (or if you want to be very clean, redirects) the results channels in `get_analysis_results`.

PR #284, opened 2022-06-09
Poetry

hartytp:
Add poetry-based workflow.

I found this really helpful for making it easy to format the code, run tests, etc without having access to the OITG internal conda packages.

I'd be happy to change the CI to be based on the poetry workflow if there is interest? The advantage I see from that is not having to maintain a docker image just for testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,42 @@
+[tool.poetry]
+name = "ndscan"
+version = "v0.2.0"
+description = "N-dimensional scans for ARTIQ"
+authors = ["David Nadlinger <david.nadlinger@physics.ox.ac.uk>"]
+license = "LGPLv3+"
+
+[tool.poetry.dependencies]
+python = "~3.8"
+numpy = "^1.22.4"
+h5py = "^3.7.0"
+pyqtgraph = "^0.12.4"
+PyQt5 = "^5.15.6"
+PyQt5-Qt5 = "^5.15.2"
+PyQt5-sip = "^12.10.1"
+qasync = "^0.23.0"
+artiq = {git = "https://github.com/m-labs/artiq.git"}
+sipyco = {git = "https://github.com/m-labs/sipyco.git"}
+oitg = {git = "https://github.com/oxfordiontrapgroup/oitg.git"}
+
+[tool.poetry.dev-dependencies]
+unittest = "^0.0"
+yapf = "^0.32.0"
+flake8 = "^4.0.1"
+toml = "^0.10.2"
+
+[tool.poetry-dynamic-versioning]
+enable = true
+vcs = "git"
+style = "pep440"
+format-jinja = "{{ base }}.{{ distance }}+g{{ commit }}"
+pattern = "^(?P<base>\\d+(\\.\\d+)*)"
+
+[build-system]
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry.core.masonry.api"
+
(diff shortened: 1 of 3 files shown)


hartytp:
@pmldrmota am I okay to merge this? I found it helpful when running tests etc since we don't have access to the conda packages.

pmldrmota:
From my side sure, I see no reason why not.

dnadlinger:
Please don't.

hartytp:
> Please don't without further discussion.

okay, I'll wait for feedback. It's not something I feel hugely strongly about, but I do think it's nice to have a way of testing things which doesn't rely on the oitg private conda packages. FWIW the config here is intended to be directly equivalent to what the CI currently does.

pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[tool.poetry]
+name = "ndscan"
+version = "v0.2.0"
+description = "N-dimensional scans for ARTIQ"
+authors = ["David Nadlinger <david.nadlinger@physics.ox.ac.uk>"]
+license = "LGPLv3+"
```
dnadlinger, line 6: Should be GPLv3.

dnadlinger:
Thanks for the wait, just saw this pop up yesterday, but didn't have the time to immediately comment.

I agree with your points in principle, although I'll add that I'm not using Conda for working on ndscan either, and installing it manually in a venv works just fine. Tests use the standard Python setup, so no Conda dependency there either. Also, for actually testing experiment execution – such as the examples, which are good to verify after changes, even though there isn't an automated toolchain yet – one still needs a patched ARTIQ at this point, which means that the Poetry install wouldn't be functional, sort of defeating the point.

Right now, I am not personally using Poetry, and it isn't exactly ubiquitous, certainly not compared to Conda (but we all know about the technical merits of that all too well). If Poetry really can be installed seamlessly in practice, I'm not in principle opposed to moving everything there, as it could provide a neatly streamliend solution for new contributors. I guess I'm finding it hard to gauge whether the typical student/post-doc would shy away from more, having to install unfamiliar tools first vs. dealing with the known setup.py/pip workflow but manually having to take care of the right dependency versions.

Either way, if we do end up adding Poetry files into the source tree, the one thing I want to make sure is that we have a clear strategy for packaging going forward. From expecience, having to maintain multiple build systems in parallel that one doesn't even use can be exceedingly draining (but if they drift out of sync, this will leave a bad impression of the project quality). Can Nix deal with pyproject.toml files that reference Poetry? (Seems like the dependencies not being specified in `[build-system] requires = […]`, but only the Poetry-specific sections might potentially be an issue?) And at least until somebody rejigs the Oxford uni group's internal CI pipeline, we'll have to continue supporting Conda packages for Windows distribution. `conda/` should be removed from this repo anyway, but can Conda consume a Poetry `pyproject.toml`?

hartytp:
> although I'll add that I'm not using Conda for working on ndscan either, and installing it manually in a venv works just fine

Noted. In practice it's a non-trivial amount of work to set the venv up to the point where one can reproduce what the CI server will do in advance of submitting a PR. The `toml` file here is essentially just a basic recipe for that.

> although I'll add that I'm not using Conda for working on ndscan either, and installing it manually in a venv works just fine

It's still better than nothing, but I do take the point. I've been chewing over the idea of publishing an artiq package on pypi which has the ndscan patch applied. This wouldn't be much work and would make it easy for people to play around with ndscan.

Anyway, it sounds like this is probably not going to get merged so I'll close the PR. At least the code is there for next time I want to make a PR.

PR #307, opened 2022-08-10
Recompute parameter defaults in continuous scans

pmldrmota:


dnadlinger:
Would be good to extend the test coverage here.

PR #310, opened 2022-09-27
Potential bugfix - _default_params currently grows infinitely

charlesbaynham:
Hi @dnadlinger,

I've opened this PR not because it should be merged, but because I wanted to highlight what I think may be a bug, and ask you if it is. 

At the moment, it seems like calling `run_once()` more than once results in repeated calls to `init_params()` which then results in `Fragment._default_params` having another entry appended to it. I've added a unit test in this PR to show the problem. 

Do you consider this a bug? If not, what's the recommended way of calling an ExpFragment's `run_once` functionality from an `EnvExperiment` more than once in a session?

PS thanks for the library, it's great :)

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -573,6 +573,11 @@ class Fragment(HasEnvironment):
                 if path_matches_spec(self._fragment_path, override_pathspec):
                     store = override_store
             if not store:
+                for default_param, default_store in self._default_params:
+                    if param == default_param:
+                        store = default_store
+                        break
+            if not store:
                 identity = (param.fqn, self._stringize_path())
                 try:
                     value = param.eval_default(self._get_dataset_or_set_default)
--- a/test/test_experiment_entrypoint.py
+++ b/test/test_experiment_entrypoint.py
@@ -10,7 +10,8 @@ from sipyco import pyon
 from fixtures import (AddOneFragment, ReboundAddOneFragment, TrivialKernelFragment,
                       TransitoryErrorFragment, MultiPointTransitoryErrorFragment,
                       RequestTerminationFragment, AddOneAggregate,
-                      TrivialKernelAggregate, TwoAnalysisAggregate, ReadParamDefault)
+                      TrivialKernelAggregate, TwoAnalysisAggregate, ReadParamDefault,
+                      MultiReboundAddOneFragment)
 from mock_environment import HasEnvironmentCase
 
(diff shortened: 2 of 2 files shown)


dnadlinger:
Hi @charlesbaynham, thanks for the report. To put it briefly, I just never really considered the case of calling `run_fragment_once()` multiple times (or `init_params()`, for that matter).

The `[create_and_]run_fragment_once()` methods came about to provide a simple, one-stop shop, "just do the sensible thing" way of initialising the necessary parameter/results machinery for running ExpFragments from the non-ndscan universe. In our case, we tend to only do this to instantiate parameter-less, automatically scheduled versions of certain calibration experiments, etc., where there is little control flow in the calling EnvExperiment. Thus, we just never ran into the issue you described. 

First of all, the silent, unbounded growth of `._default_params` is definitely a bug; `init_params()` should either complain when called multiple times, or do something sensible. Would you mind opening an issue for that (or proposing a fix in this PR)?

What we would currently do for cases where `run_once()` is called multiple times is likely to write some custom execution logic (or set up a scan via `TopLevelRunner`), as "entering and leaving the ndscan universe" multiple times via repeated `run_fragment_once()` calls necessarily incurs kernel compilation overhead each time (for experiments that involve the core device). It might still be worth to make `run_fragment_once()` cleanly support multiple calls and/or improve the documentation (and potentially make more of the initialisation-related API non-underscore-private).

charlesbaynham:
Alrighty, I've pushed a simple fix that just checks `_default_params` for a given parameter before it makes a new store which does the job for this scenario.

Thanks for the tip about avoiding kernel recompilation - I encountered this with code that doesn't use the core, but I'll need that eventually so the pointers are useful!

dnadlinger:
Great, thanks, and apologies for the delay. One thing to note is that this now won't recompute the default parameters from datasets anymore on repeated calls to `init_params`. I'm not sure what the cleanest behaviour is – but either way, I guess we should decide on one, and document it.

charlesbaynham:
> Great, thanks, and apologies for the delay. One thing to note is that this now won't recompute the default parameters from datasets anymore on repeated calls to `init_params`. 

True. I think that's probably the behaviour we want, otherwise why would the user be using an already-initialised fragment when they could just construct a new one? If you agree I'll add to the docs

charlesbaynham:
Hey @dnadlinger, sorry for the nudge: do you agree that this is a reasonable fix? If so, I'll write it up.

Issue #210, opened 2020-08-26
Changing unit in setattr_param_rebind does not set scale


Timboski:
Changing the `unit` in `setattr_param_rebind` changes the unit that is displayed in the UI, however it does not change the scale to match.

`self.setattr_param(...blah..., unit="us")`
`self.setattr_param_rebind(...blah..., default=100 * ms, unit="ms")`

The UI shows `100000 ms`

As a workaround, the scale must also be set:
`self.setattr_param_rebind(...blah..., default=100 * ms, unit="ms", scale=ms)`

dnadlinger:
This is a bit non-trivial to implement in the current, generic version of `setattr_param_rebind`. Patches welcome, though.

JammyL:
I think there are two options to implement while keeping things general.

1. Don't deepcopy the param. `setattr_param_like` performs `new_param = deepcopy(template_param)`. If we instead reinitialise based upon parameters in the `template_param`, then we can force the scale to check against the units again.

2. (I prefer this one) Add a `reinit` function to the params. This would perform any operations like rechecking the scale that are required for a particular `Param`. For the `StringParam` this would just `pass`.

Thoughts @dnadlinger?

dnadlinger:
I'd prefer 1), i.e. getting away from `deepcopy` and directly `setattr`-ing members in the first place. This is a pretty strong code smell; other objects shouldn't be messing around with the internal state.

JammyL:
> I'd prefer 1), i.e. getting away from `deepcopy` and directly `setattr`-ing members in the first place. This is a pretty strong code smell; other objects shouldn't be messing around with the internal state.

The benefit of keeping the `deepcopy` is that using the `Param`'s current state is very easy. Meaning that ommiting any new args such as the `unit`, `scale`, or `min`/`max`  and just using the "default" is done for you.

If we don't `deepcopy`, then either we need to depend on the `__init__` `args` being recovereable from the current `Param` object, or we store them explictly (or a third thing I haven't thought of)

dnadlinger:
> Meaning that omiting any new args such as the unit, scale, or min/max and just using the "default" is done for you.

That was indeed the reason for using the deepcopy hack in the first place, but as you found out, messing directly with the internal state of an object tends to produce weird results.

Just storing a dict of all the args in the constructor and merging that with any kwargs should be a bullet-proof, clean, easy to understand solution.

dnadlinger:
[...]

PR #315, opened 2022-11-24
Create new Param instance on rebind to set scale

JammyL:
Addresses #210 

<del/>FloatParam and IntParam recompute scale if required. 

<del/>Misses (very) edge case:
        - Param has fixed desired scale.
	- Param is rebound with new unit (but you want the old scale?)
	- Param will be adjusted to match scale of new unit
<del/>More importantly - catches not setting the scale manually and wanting to change unit

<del/>StringParam passes

Create `ParamBase` class for storing initialisation parameters
Now create new `Param` on rebind and call with stored + updated initialisation parameters.
This allows the scale for int/float params to be recomputed if required

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -363,10 +363,10 @@ class Fragment(HasEnvironment):
             "already rebound?".format(original_name))
 
         template_param = original_owner._free_params[original_name]
-        new_param = deepcopy(template_param)
-        new_param.fqn = self.fqn + "." + name
-        for k, v in kwargs.items():
-            setattr(new_param, k, v)
+        init_params = deepcopy(template_param.init_params)
+        init_params.update(kwargs)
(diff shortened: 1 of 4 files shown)


ndscan/experiment/fragment.py
```diff
@@ -367,6 +367,8 @@ def setattr_param_like(self,
         new_param.fqn = self.fqn + "." + name
         for k, v in kwargs.items():
```
JammyL, line 368: Actually, it may make sense to add these to `reinit`?

dnadlinger:
See discussion in #210 – also needs unit tests.

JammyL:
I've updated the rebind test to also change the unit. Doesn't feel worth adding an extra test

dnadlinger:
Apologies for the delay; I'm still just fundamentally unhappy with the slightly messy design here (which isn't this PR's fault).

PR #321, opened 2023-04-02
Create BoolParam parameter type

AVB25:
Create basic implementation of boolean parameters, for easy handling of boolean arguments by parent fragments and scanning over booleans (not implemented). 

UnitTests included.

In future, add scans over bool parameters and present the results in two separate graphs for `True` and `False` cases. Similarly, could implement Enum parameters.

### DIFF
--- a/ndscan/experiment/parameters.py
+++ b/ndscan/experiment/parameters.py
@@ -13,12 +13,17 @@ from artiq.language import units
 from typing import Any, Dict, Optional, Tuple, Type, Union
 from ..utils import eval_param_default, GetDataset
 
-__all__ = ["FloatParam", "IntParam", "StringParam"]
+__all__ = ["FloatParam", "IntParam", "StringParam", "BoolParam"]
 
 
 def type_string_to_param(name: str):
     """Resolve a param schema type string to the corresponding Param implementation."""
-    return {"float": FloatParam, "int": IntParam, "string": StringParam}[name]
+    return {
+        "float": FloatParam,
+        "int": IntParam,
+        "string": StringParam,
+        "bool": BoolParam
+    }[name]
 
 
 class InvalidDefaultError(ValueError):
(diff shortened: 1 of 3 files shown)


ndscan/experiment/parameters.py
```diff
@@ -374,3 +417,36 @@ def eval_default(self, get_dataset: GetDataset) -> str:
...
+class BoolParam:
+    HandleType = BoolParamHandle
+    StoreType = BoolParamStore
+    CompilerType = TBool
+
+    def __init__(self,
+                 fqn: str,
+                 description: str,
+                 default: bool,
+                 is_scannable: bool = False):
```
dnadlinger, line 431: This should probably default to `True`, as the idea for `is_scannable` is to be able to disable scans for a few parameters where one wants to make absolutely sure they are not scanned by accident, or the rare cases where some code can't handle scans.

dnadlinger:
Thanks a lot, looks good! Could you address the one comment, and then perhaps fold the formatting commit/the extra change into the original three commits to keep the history easy to access? (If you haven't worked with rebasing/… before, we can discuss on Thursday.)

dnadlinger:
Thanks a lot, looks good! Could you address the one comment, and then perhaps fold the formatting commit/the extra change into the original three commits to keep the history easy to access? (If you haven't worked with rebasing/… before, we can discuss on Thursday.)

dnadlinger:
Thanks a lot, looks good! Could you address the one comment, and then perhaps fold the formatting commit/the extra change into the original three commits to keep the history easy to access? (If you haven't worked with rebasing/… before, we can discuss on Thursday.)

dnadlinger:
We should totally look into which parts of all that duplicated code we can remove now that the ARTIQ compiler is quite a bit more stable overall.

ndscan/experiment/parameters.py
AVB25: Sure! I had defaulted it to `False` as the param doesn't currently handle scans, but I'll change that.

PR #323, opened 2023-04-21
Stacked plots

JammyL:
Brings in functionality from OI to have vertically stacked plots instead of all on the same axis.
Would be great to follow up with some sort of flag that allows the user to choose if two datasets should be split or share axes.

image

### DIFF
--- a/ndscan/plots/container_widgets.py
+++ b/ndscan/plots/container_widgets.py
@@ -103,7 +103,7 @@ class PlotContainerWidget(QtWidgets.QWidget):
         self._alternate_plots["main plot"] = self.plot
         self.plot.error.connect(self._show_error)
         self.plot.ready.connect(lambda: self._show(self.plot))
-        self.plot.alternate_plot_requested.connect(self._show_alternate_plot)
+        # self.plot.alternate_plot_requested.connect(self._show_alternate_plot)
 
         # All subscan roots, to keep the objects alive.
         self._subscan_roots = None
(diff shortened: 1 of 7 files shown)


ndscan/plots/container_widgets.py
```diff
@@ -103,7 +103,6 @@ def __init__(self, model: Model):
         self._alternate_plots["main plot"] = self.plot
         self.plot.error.connect(self._show_error)
         self.plot.ready.connect(lambda: self._show(self.plot))
-        self.plot.alternate_plot_requested.connect(self._show_alternate_plot)
```
dnadlinger, line 106: What's this about?

ndscan/plots/xy_1d.py
```diff
@@ -298,11 +296,9 @@ def _point_clicked(self, scatter_plot_item, spot_items):
         # overlap; the user can always zoom in if that is undesired.
         spot = spot_items[0]
         self._highlight_spot(spot)
-        self.selected_point_model.set_source_index(spot.index())
```
dnadlinger, line 301: This would break the subscan plots.

dnadlinger:
@pmldrmota and I spent a good while today cleaning up the remaining issues (subscans, layouting, etc.) and a first version is now merged as https://github.com/OxfordIonTrapGroup/ndscan/commit/2a7b13e835e19706062b65d2087226b3742e433c – thanks all!

PR #324, opened 2023-04-21
Fix bug in BoolParam

AVB25:
Fix bug in BoolParam.eval_default() where the default value wasn't properly computed.

### DIFF
--- a/test/test_experiment_parameters.py
+++ b/test/test_experiment_parameters.py
@@ -21,6 +21,16 @@ class FloatParamCase(unittest.TestCase):
                 "type": "float"
             })
 
+    def test_evaluate_default(self):
+        def mock_get_dataset(key: str, default=None):
+            return {"baz": 42.}[key]
+
+        param = FloatParam("foo", "bar", 0.)
+        self.assertEqual(param.eval_default(mock_get_dataset), 0.)
+
+        param = FloatParam("foo", "bar", "dataset('baz', 0.)")
+        self.assertEqual(param.eval_default(mock_get_dataset), 42.)
+
 
 class IntParamCase(unittest.TestCase):
     def test_describe(self):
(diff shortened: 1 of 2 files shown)


test/test_experiment_parameters.py
```diff
@@ -21,6 +21,21 @@ def test_describe(self):
                 "type": "float"
             })
 
+    def test_evaluate_default(self):
+        def placeholder_get_default(key, default):
+            return key
+
+        param_float_default = IntParam("foo", "bar", 0., min=-1, max=1, unit="baz", scale=1)
```
dnadlinger, line 28: This should be `FloatParam`, presumably. I'd also remove all the `min`/`max`/`unit`/`scale` to put the focus on what is acutally being tested here (the default value).

test/test_experiment_parameters.py
```diff
@@ -21,6 +21,21 @@ def test_describe(self):
                 "type": "float"
             })
 
+    def test_evaluate_default(self):
+        def placeholder_get_default(key, default):
```
dnadlinger, line 25: As per the parameter name, this function is supposed to be `get_dataset`, not `get_default`. The first parameter is also supposed ot be a string as per the type annotations (although this is not currently enforced at execution time), so the tests are weirdly backwards.

Maybe something like
```py
def mock_get_dataset(key, default=None):
  return {"baz": 42}[key]
```
which would also check that the key is passed correctly, and below `IntParam("foo", "bar", "dataset('baz')")`?

test/test_experiment_parameters.py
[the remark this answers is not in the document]
AVB25: Implemented in latest commit

test/test_experiment_parameters.py
AVB25: Other than the previous comment, implemented in previous commit.

dnadlinger: I was just referring to the dataset() argument order in the previous tests.

dnadlinger (approved):
Looks good to me – could you please rebase ok top of latest master? The CI checks won't allow this to be merged otherwise.

dnadlinger:
Ah, GitHub now has a button for this – just did it myself.

PR #325, opened 2023-05-02
WIP: Add live analysis

JammyL:
### Code needs tidying ~~and removing a bunch of copy pastes!~~

~~Add LiveAnalysis~~

~~This is similar to DefaultAnalysis, but with a stripped back implementation to handle only reading result channels and pushing to other datasets~~

Adds LiveAnalysisRunner, which handles the live_analysis passed from `ExpFragment.get_live_analyses`
The required CustomAnalysis classes are determined at build time through similar filters 

A change that is worth discussion, at run time the `dataset_prefix` is passed from the `TopLevelRunner` to the fragment. This allows the fragment to access datasets at run time which is very useful for the live analysis.

~~Most of the changes to `Fragment` stem from  the `ScanRunner` not having access to the `TopLevelRunner`. For the `DefaultAnalysis` this is run after the `ScanRunner` is done and so isn't an issue.~~

~~We could pass the `TopLevelRunner` as an argument to `ScanRunner.run` and then call the `live_analysis` on `TopLevelRunner`.~~

Alternatively (and perhaps most sensibly) we pass the `LiveAnalysis` objects down to the `ScanRunner` and have `ScanRunner` handle them. **- Have done this through the LiveAnalysisRunner**

Of course, this is all assuming we want to stick with a model roughly like the `LiveAnalysis`

### DIFF
--- a/ndscan/experiment/default_analysis.py
+++ b/ndscan/experiment/default_analysis.py
@@ -249,29 +249,6 @@ class CustomAnalysis(DefaultAnalysis):
         return [a.describe(context) for a in annotations]
 
 
-class LiveAnalysis(CustomAnalysis):
-    def execute(self,
-                axis_data: Dict[AxisIdentity, list],
-                result_data: Dict[ResultChannel, list],
-                ) -> List[Dict[str, Any]]:
-        user_axis_data = {}
-        for handle in self._required_axis_handles:
-            user_axis_data[handle] = axis_data[handle._store.identity]
-
-        try:
-            self._analyze_fn(user_axis_data,
-                             result_data,
-                             self._result_channels)
-        except TypeError as orignal_exception:
-            # Tolerate old analysis functions that do not take analysis result channels.
-            try:
-                self._analyze_fn(user_axis_data, result_data)
-            except TypeError:
-                # KLUDGE: If that also fails (e.g. there is a TypeError in the actual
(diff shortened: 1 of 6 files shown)


JammyL:
This currently breaks subscans - I need to pass the result channel sinks to the scanrunner in the subscan implementation.
However, we're currently considering how to change this anyway with on-kernel subscans

JammyL:
> Could you provide a bit of background about what problem you're looking to solve here?

Sure, I think this has been burried in some private conversations, so definitely worth explaining here too.
I've implenented an ndscan-ified version of our RBM scripts in our experiment specific repository. The current implenetation of this is a 2D scan over `no_cliffords` and `seed` (for generating a random sequence). These then run in a random order. I then average over the results from different `seed` values at each `no_cliffords` after each data point is taken and push to another dataset.
Then another plot can subscribe to these datasets.

An alternative to the above would be to do a series of subscans, but this lacks random order that a 2D scan can provide, and so visualising a RBM run would be slow

This PR is to provide a place for discussion at how this kind of repeated analysis could look to enable such an experiment.
A possible extra usecase I think we could work into this is gettting live data for subscans so we don't have to wait until the end of each point.

hartytp:
Thanks for explaining!

dnadlinger:
Closing, with our internal design notes now at #388, #389, #390.

PR #328, opened 2023-06-21
ndscan_to_txt

pmldrmota:
I found myself often copying the same lines of code to write raw data to text files for analysis or plotting in other programs.

@dnadlinger: As mentioned in the FIXME comment, I am not sure when there would be more than one prefix, and which one to use in this case. The current code would fail if `len(prefixes)>1`.

### DIFF
--- a/ndscan/to_txt.py
+++ b/ndscan/to_txt.py
@@ -1,0 +1,98 @@
+"""Standalone tool to write scan data to text files"""
+
+import argparse
+import os
+import json
+import numpy as np
+
+from .show import load_h5
+from .results.tools import get_source_id
+from .utils import strip_suffix, shorten_to_unambiguous_suffixes
+
+
+def get_argparser():
+    parser = argparse.ArgumentParser(
+        description="Extracts ndscan results from ARTIQ HDF5 file to text file",
+        epilog=(
+            "Instead of a file name, just a run id or 'magic' source string can be "
+            "supplied, which is then resolved using oitg.results (e.g. 'alice_12345', "
+            "or just '12345' to infer the experiment name from the environment)."))
+    parser.add_argument("--prefix",
+                        default=None,
+                        type=str,
+                        help="Prefix of root in dataset tree (default: auto-detect)")
+    parser.add_argument("path",
+                        metavar="FILE",
+                        type=str,
+                        help="Path to HDF5 results file")
+    return parser
+
+
+def main():
+    args = get_argparser().parse_args()
+
+    path, datasets, prefixes, schema = load_h5(args)
+    if len(prefixes) > 1:
+        raise Exception("More than one ndscan prefix found. Please specify one. " +
+                        f"Prefixes: {prefixes}")
+    prefix = prefixes[0]
+
+    # Use parameter names from last part of FQNs (maybe more if ambiguous) as column
+    # labels in the header line.
+    # FIXME: This breaks if there are two axes with the same FQN (but different paths).
+    fqns = [ax["param"]["fqn"] for ax in json.loads(datasets[prefix + "axes"][()])]
+    axes_names = list(
+        shorten_to_unambiguous_suffixes(
+            fqns, lambda fqn, n: ".".join(fqn.split(".")[-n:])).values())
+
+    channel_names = []
+    point_data = {}
+
+    channel_schemata = json.loads(datasets[prefix + "channels"][()])
+
+    for i, name in enumerate(axes_names):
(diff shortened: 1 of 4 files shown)


dnadlinger:
Hmm, I'm a bit unsure what the right tradeoffs are here.

First of all, though, I'm all for adding more analysis tooling to ndscan. This is a hugely under-exploited part of the design; having rich metadata available in the results files for standardised analysis components was a major factor in the design, and we are currently not making much use of it.

As for the specifics, though:

 - Multiple prefixes referes to multiple `TopLevelRunner`s/… generating independent output in the datasets (internal example: the micromotion compensation servo). `ndscan_show` allows you to switch between them using the context menu.
 - `ndscan_show` doesn't seem to be a particularly natural place to put this, as it is entirely independent of the UI plotting functionality. What about a little `ndscan_to_txt` command line tool instead?
 - If we do want to add export functionality to `ndscan_show`, the more straightforward way to go seems to me to fix pyqtgraph's built-in data export integration, or otherwise implement something similar, where one can manually export the data currently on screen, even if that's e.g. a subscan buried a few layers deep.
 - Generally, I think in the scientific context, data provenance is an important, underappreciated aspect, and ndscan should encourage good practices wherever possible (like makign sure the rid is shown in plot screen clippings, etc.). To this end, it might actually be preferrable to have as much analysis code as possible operate directly on the source HDF5 files, as they are the "one source of truth" in the system, and contain ancillary metadata (time/Git revision/…). If we do want to provide textual export facilities, I wonder whether we should make it emit a richer header with (at least) the original rid and system id as well.
 - Sorting the data rows this way loses information (the acquisition order), which can be important to detect drifts/drop-outs/etc. If we want to sort the data by default, what about adding an additional point index column to preserve this information?

setup.py
```diff
@@ -9,7 +9,8 @@
     author="David Nadlinger",
     packages=find_packages(),
     entry_points={
-        "console_scripts": ["ndscan_dataset_janitor = ndscan.dataset_janitor:main"],
+        "console_scripts": ["ndscan_dataset_janitor = ndscan.dataset_janitor:main",
+                            "ndscan_to_txt = ndscan.to_text:main"],
```
dnadlinger, line 13: `to_txt`

ndscan/to_txt.py
```diff
@@ -0,0 +1,89 @@
...
+    column_names = axes_names + channel_names
+    ordered_data = [point_data[name][order] for name in column_names]
+
+    # Save into location where this script is run from.
+    target_path = strip_suffix(os.path.basename(path), ".h5") + ".dat"
+    print(f"Saving axes and point data to '{target_path}'")
+    with open(target_path, 'wb') as f:
+        f.write(bytes(f"# Source ID: {get_source_id(datasets, prefixes)}\n", "UTF-8"))
+        f.write(bytes(f"# Source prefix: {prefix[:-1]}\n", "UTF-8"))
+        f.write(bytes(f"# Source HDF5: {path}\n", "UTF-8"))
```
dnadlinger, line 80: I guess NumPy just ignores these lines? I wonder whether we should just stick everything on one (and stick to that in the future as well) to make it easier to skip from other tools.

ndscan/to_txt.py
```diff
@@ -0,0 +1,89 @@
...
+    args = get_argparser().parse_args()
+
+    path, datasets, prefixes, schema = load_h5(args)
+    if len(prefixes) > 1:
+        raise Exception("More than one ndscan prefix found. Please specify one. " +
+                        f"Prefixes: {prefixes}")
+    prefix = prefixes[0]
+
+    axes = json.loads(datasets[prefix + "axes"][()])
+    axes_names = [ax["param"]["fqn"].split(".")[-1] for ax in axes]
```
dnadlinger, line 41: Hmm, there could be two axes with the same parameter (e.g. a 2D scan of RSB π time vs carrier π time) – see ndscan.utils.shorten_to_unambiguous_suffixes`. For the same reason, the path (instead of fqn) would probably be more approrpriate, though the situation is a bit more complex for scans (overrides in general), as one can use the glob (`*`) matches. I guess we can live with this for now.

dnadlinger:
Feel free to merge after addressing the typo and (if you agree) combining the headers into one line.

I wonder, though, whether this (space-delimted?) format or CSV is more universally supported.

pmldrmota:
For me, space-delimited is the most useful because it works directly with pgfplots and Mathematica.
If someone wants a different format we can always add those options.

Issue #101, opened 2019-05-22
plots: Colour bar for 2D scans


RHanley1:
When a 2D scan is performed, a color-map is used to display the data. Is it possible to add a scale bar to the applet so that one can read the data values before post-processing?

dnadlinger:
Would be very, very useful, yes; just like #102.

This should be fairly easy to add, as only changes to https://github.com/OxfordIonTrapGroup/ndscan/blob/master/ndscan/plots/image_2d.py should be necessary.

NearConstant:
I'm also having this issue, was there any progress on it?

dnadlinger:
Fixed in #329, result channel label added in https://github.com/OxfordIonTrapGroup/ndscan/commit/b8f12b40aca7b6a0cf19801a1838d20aa4d278bb.

PR #329, opened 2023-06-21
Add colorbar to 2D plots!

JammyL:
Addresses #101 
Adds colorbar to Image2DPlotWidget and _ImagePlot
No longer scale z values or map to RGBA

Compare plot with master checkout to this branch to make sure plots aren't disrupted at all
Master:
master_checkout

This branch: 
colorbar_v2

### DIFF
--- a/ndscan/plots/image_2d.py
+++ b/ndscan/plots/image_2d.py
@@ -53,7 +53,7 @@ def _coords_to_indices(coords, range_spec):
 
 
 class _ImagePlot:
-    def __init__(self, image_item: pyqtgraph.ImageItem, active_channel_name: str,
+    def __init__(self, image_item: pyqtgraph.ImageItem, colorbar: pyqtgraph.ColorBarItem, active_channel_name: str,
                  x_min: Union[float, None], x_max: Union[float, None],
                  x_increment: Union[float,
                                     None], y_min: Union[float,
@@ -61,6 +61,7 @@ class _ImagePlot:
                                                                             None],
                  y_increment: Union[float, None], hints_for_channels: Dict[str, dict]):
         self.image_item = image_item
+        self.colorbar = colorbar
         self.active_channel_name = active_channel_name
         self.hints_for_channels = hints_for_channels
 
@@ -136,9 +137,7 @@ class _ImagePlot:
 
             # TODO: Splat old data for progressively less blurry look on refining scans?
             self.image_data = np.full(
-                (_num_points_in_range(x_range), _num_points_in_range(y_range), 4),
-                0,
-                dtype="ubyte")
+                (_num_points_in_range(x_range), _num_points_in_range(y_range)), 0.0)
 
             self.image_rect = QtCore.QRectF(
                 QtCore.QPointF(x_range[0] - x_range[2] / 2,
@@ -151,14 +150,17 @@ class _ImagePlot:
         x_inds = _coords_to_indices(x_data[num_skip:num_to_show], self.x_range)
         y_inds = _coords_to_indices(y_data[num_skip:num_to_show], self.y_range)
 
-        z_min, z_max = self.current_z_limits
-        z_scaled = (z_data[num_skip:num_to_show] - z_min) / (z_max - z_min)
-
         cmap = colormaps.plasma
         if self._get_display_hints().get("coordinate_type", "") == "cyclic":
             cmap = colormaps.kovesi_c8
-        self.image_data[x_inds, y_inds, :] = cmap.map(z_scaled)
 
+        for x, y, z in zip(x_inds, y_inds, z_data):
+            self.image_data[x, y] = z
+
+        self.colorbar.setColorMap(cmap)
+        self.image_item.setColorMap(cmap)
+        self.colorbar.setLevels(self.current_z_limits)
+        self.image_item.setLevels(self.current_z_limits)
         self.image_item.setImage(self.image_data, autoLevels=False)
         if num_skip == 0:
             # Image size has changed, set plot item size accordingly.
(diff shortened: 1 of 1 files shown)


JammyL:
> Currently encountering (not much of) an edge case. Since this no longer uses the ARGB mapping, undefined points are rendered as 0.0. This will take some work to iron out since I (as far as I can find) can't use the ARGB value without messing up the scale on the colorbar, nor can I edit the ticks on the colorbar to match the scale. The latter feels like it should be a core part of the pyqtgraph functionality, but not that I've found.

Changed default in array to `np.NaN` and this is now fixed :)

JammyL:
This requires `pyqtgraph==0.13.3`. `0.12` doesn't have `ImagePlotItem.addColorBar`

ndscan/plots/image_2d.py
```diff
@@ -137,7 +138,7 @@ def _update(self):
 
             # TODO: Splat old data for progressively less blurry look on refining scans?
             self.image_data = np.full(
-                (_num_points_in_range(x_range), _num_points_in_range(y_range)), 0.0)
+                (_num_points_in_range(x_range), _num_points_in_range(y_range)), np.NaN)
```
dnadlinger, line 141: I didn't know the camel-case version even existed! To cite https://numpy.org/doc/stable/reference/constants.html#numpy.NaN, "use `nan`".

ndscan/plots/image_2d.py
```diff
@@ -151,14 +151,17 @@ def _update(self):
...
-        z_min, z_max = self.current_z_limits
-        z_scaled = (z_data[num_skip:num_to_show] - z_min) / (z_max - z_min)
-
         cmap = colormaps.plasma
         if self._get_display_hints().get("coordinate_type", "") == "cyclic":
             cmap = colormaps.kovesi_c8
-        self.image_data[x_inds, y_inds, :] = cmap.map(z_scaled)
 
+        for x, y, z in zip(x_inds, y_inds, z_data):
+            self.image_data[x, y] = z
```
dnadlinger, line 159: This could be moved up now; it does not rely on the colour map anymore.

ndscan/plots/image_2d.py
```diff
@@ -151,14 +151,17 @@ def _update(self):
...
+
+        self.colorbar.setColorMap(cmap)
+        self.image_item.setColorMap(cmap)
```
dnadlinger, line 162: Presumably, this could be moved into `_initialise_series` now instead of re-setting the colour map every time?

ndscan/plots/image_2d.py
```diff
@@ -227,8 +230,13 @@ def bounds(schema):
 
         image_item = pyqtgraph.ImageItem()
         self.addItem(image_item)
-        self.plot = _ImagePlot(image_item, self.data_names[0], *bounds(self.x_schema),
-                               *bounds(self.y_schema), hints_for_channels)
+        colorbar = self.getPlotItem().addColorBar(image_item,
+                                                  colorMap=colormaps.plasma,
+                                                  values=(0, 1),
```
dnadlinger, line 235: What is the significance of passing `(0, 1)` here? (I assume this is overridden using `setLevels` later?)

ndscan/plots/image_2d.py
JammyL: Yep, so I'll resolve this. Correct me if I've misunderstood your comment

JammyL: Actually, after checking - the `self.image_item.setColorMap(cmap` can be remove, it inherits from the `colorbar` anyway

ndscan/plots/image_2d.py
JammyL: No significance, and we are overriding, so will remove. In fact, the same applies to the `colorMap=...`

ndscan/plots/image_2d.py
JammyL: and the same applies for `self.image_item.setLevels`!

dnadlinger:
> This requires pyqtgraph==0.13.3. 0.12 doesn't have ImagePlotItem.addColorBar

Can we add this to `conda/…` to make sure our internal deployments work (probably in the form of a `>=` constraint rather than `==`)?

Perhaps actuallys pecifying the dependencies in setup.py is worth another go too, it's clearly the right thing to do…

JammyL:
> > This requires pyqtgraph==0.13.3. 0.12 doesn't have ImagePlotItem.addColorBar
> 
> Can we add this to `conda/…` to make sure our internal deployments work (probably in the form of a `>=` constraint rather than `==`)?
> 
> Perhaps actuallys pecifying the dependencies in setup.py is worth another go too, it's clearly the right thing to do…

I agree, I think setup.py could use some proper dependencies. Which should then also work for conda.
On the linux side of things - how do we approach this for linux (where we are using [nix-oitg](https://github.com/OxfordIonTrapGroup/nix-oitg))? If ndscan is installed via pip after, the setup.py will take care of this.

To get this PR over the line, I'd suggest we add just this `pyqtgraph==0.13.3`  to `setup.py` for now. This is with the intention that we chase up some issues in our environments - most importantly that our `artiq.__version__` is not `PEP440` compliant.

dnadlinger:
Agreed – but why force exactly 0.13.3? It should presumably continue to work on newer versions (and, incidentally, `addColorBar` appears to exist since 0.12.4).

If we want to pin a specific version for whatever other reason, doing this in the "top-level application" context (here: artiq-env) seems more appropriate.

JammyL:
You're right, I was meant to put `>=0.13.3`. Though as you've found, we can do `0.12.4`

JammyL:
Tested with `pyqrgraph==0.13.3` and `==0.12.4`. Have put `>=0.12.4` in `setup.py`

ndscan/plots/image_2d.py
```diff
@@ -150,14 +150,14 @@ def _update(self):
 
         x_inds = _coords_to_indices(x_data[num_skip:num_to_show], self.x_range)
         y_inds = _coords_to_indices(y_data[num_skip:num_to_show], self.y_range)
-
-        z_min, z_max = self.current_z_limits
-        z_scaled = (z_data[num_skip:num_to_show] - z_min) / (z_max - z_min)
+        for x, y, z in zip(x_inds, y_inds, z_data):
```
dnadlinger, line 153: Seems like this should also take `z_data[num_skip:num_to_show]`?

JammyL: Ah yes, whoops!

PR #331, opened 2023-07-10
Annotation docs

william-cutler:
This information was rather hard to find, so hopefully this documentation clears some of it up. I found things out by looking at the code, specifically xy_1d.py and how it uses the Annotation, but I likely missed some things.

### DIFF
--- a/ndscan/experiment/default_analysis.py
+++ b/ndscan/experiment/default_analysis.py
@@ -88,6 +88,39 @@ class Annotation:
                  coordinates: Optional[dict] = None,
                  parameters: Optional[dict] = None,
                  data: Optional[dict] = None):
+        """Specifies the annotation to be plotted. This documentation is likely very
+	        incomplete due to use of multiple strings in specification.
+	        Args:
+	            kind (str): One of 'location', 'curve', or 'computed_curve'.
+	                See ndscan.plots.xy_1d.XY1DPlotWidget._update_annotations for usage.
+	            coordinates (Optional[dict], optional):
+	                * For kind == 'location', should be a single entry, key is a
+	                ParamHandle value is a single number for the location.
+	                * For kind == 'curve', should be a dict with 2 entries. Keys are the x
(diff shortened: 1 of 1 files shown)


dnadlinger:
Hi – thanks for the PR! It is true that this is only documented implicitly via the `rabi_flop_fit` example, so it would indeed be a good idea to make this more discoverable.

Generally, `ndscan` uses Sphinx to build API docs from reStructuredText-style comments, so please stick with that (`:param …:` blocks, double backquotes for inline code symbols, cross-reference syntax, etc.). I've now added an explicit mention to `docs/coding_conventions.rst`, which gets published as https://oxfordiontrapgroup.github.io/ndscan/coding-conventions.html. By default, Sphinx only reads the class-level docstring for everything (including the constructor parameters); the additional one for `__init__` you've added won't be taken into account.

Even after adapting the formatting syntax, this would probably still be a little bit hard to parse. If we do want to document everything here, perhaps rendering this as a table would be cleaner (e.g. with columns for `kind` and the other arguments, and then simply `-` for the fields that don't apply?). There is an example on how to do that in reStructuredText here: https://oxfordiontrapgroup.github.io/ndscan/apidocs.html#ndscan.experiment.result_channels.ResultChannel

As mentioned on Slack, though, I wonder whether it wouldn't be considerably more user-friendly to simply provide a specialised API for the different annotation types. One way could be to make this the first instance of a dedicated model layer that is converted from/to JSON on both the experiment and applet/results analysis sides of the code as necessary (as Tom and others suggested elsewhere). Another, probably much simpler way would be to simply provide a factory function for each kind of annotation that can then take only the necessary parameters, and also provide better documentation.

dnadlinger:
Implemented the approach of documenting factory functions in https://github.com/OxfordIonTrapGroup/ndscan/pull/397, see https://oxfordiontrapgroup.github.io/ndscan/apidocs.html#module-ndscan.experiment.annotations. Thanks again for kicking this off last year!

PR #337, opened 2023-09-21
1D applet: average points with same coordinate

pmldrmota:
When running a scan with `repeats > 1`, plots can quickly become crowded.
This creates a new feature, where overlapping points are averaged and their error bars combined.
The context menu provides the option to switch between the old view and the averaged view.
If data points already have error bars, those will be taken into account, as well as the spread of the overlapping points.
If there are no error bars on individual points, the averaged view displays their standard deviation.

For future adaptive sampling approaches, where one wants to resample points with insufficiently low errors, this should improve the user experience.

The behaviour in the context of subscans (what happens when clicking averaged points etc) has not been tested yet (will do this soon).

### DIFF
--- a/ndscan/plots/xy_1d.py
+++ b/ndscan/plots/xy_1d.py
@@ -1,6 +1,8 @@
 import logging
 import numpy as np
 import pyqtgraph
+from collections import defaultdict
+from typing import NamedTuple
 
 from .._qt import QtCore
 from .annotation_items import ComputedCurveItem, CurveItem, VLineItem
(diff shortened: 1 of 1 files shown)


dnadlinger:
I'd prefer keeping averaging off by default, on two grounds:
 a) ndscan ought to show the data in an unadulterated fashion by default, and
 b) there is the subscan selection issue.

Also, making it a checkable "enable …" entry in the context menu rather than a "toggle" command would both be more informative, and also fit in better with the various pyqtgraph options (which are also stateful entries, rather than toggle commands).

ndscan/plots/xy_1d.py
```diff
@@ -29,54 +44,119 @@ def __init__(self, view_box, data_name, data_item, error_bar_name, error_bar_ite
...
+        y_data = np.array([np.mean(self._y_collection[x]) for x in x_data])
+
+        if new_y_err is not None:
+            y_err = np.array([
+                combined_uncertainty(self._y_collection[x], self._err_collection[x])
+                for x in x_data
+            ])
+        else:
+            y_err = np.array([
+                np.std(self._y_collection[x]) / np.sqrt(len(self._y_collection[x]))
```
dnadlinger, line 147: Shouldn't the divisor be `n - 1` here? (Possibly with some extra Student-t-weighting, even.

pmldrmota:
I changed the standard deviation and the context menu to only appear when averaging even applies and instead of "toggle", it switches between "enable" and "disable" now.

Moreover, [b5f5c3d](https://github.com/OxfordIonTrapGroup/ndscan/pull/337/commits/b5f5c3ddd118a8da288ee7d958990fd29af9b333) fixes subscans such that the most recent subscan from the ensemble making up the clicked averaged point is shown.

dnadlinger:
This was supposed to "just work"; let me work out a simpler fix.

dnadlinger:
Hmm, on second thought, perhaps the complexity of adding/removing the error bar item is not too bad. Genereally, I tried to avoid hyperoptimising everything for handling all possible state deltas in the most efficient way in ndscan. Rather, I'm usually content with just making sure "fast-forward" updates when new data comes in are enacted efficiently, and just rebuild the entire series from scratch when there is a more invasive change (such as here the enabling of averaging). The latter only happens after manual user intervention anyway, so no danger of hundreds of them per second there. This might just be acceptable, though.

Issue #102, opened 2019-05-22
plots: Cursors should display values along all axes


dnadlinger:
For 1D plots with multiple y axes, and 2D plots in general, the crosshair cursor doesn't currently display all the values.

dnadlinger:
(For 2D plots, the currently-hovered z value should also be shown.)

PR #353, opened 2023-10-06
131 plot unitsscale for shared axes channels broken

### LINKED ISSUE #131: plot: Units/scale for shared axes channels broken
See e.g. bob_103629 – even if channels on shared axes have same unit, displayed values are not correctly folded (and cursor shows raw value).

pmldrmota:
Fixes #131.
For every axis, each `(unit, scale)` combination gets a crosshair label entry with the color of the first data series of that kind.
When an axis is shared between data of different `(unit, scale)` combination, the first data series determines the scale of the axis and the axis label of all other data series gets an additional multiplier, such as `y × 100 / us`.
In the future, this could be improved by converting between different scales of different units if the multiplier is a factor of 10. The current solution works for any scale ratio, where up to 6 significant figures will be shown in the axis label.

Example plot:
image

### DIFF
--- a/ndscan/plots/cursor.py
+++ b/ndscan/plots/cursor.py
@@ -5,12 +5,12 @@ from .utils import SERIES_COLORS
 
 
 class CrosshairLabel(pyqtgraph.TextItem):
+    """Text item to be displayed alongside the cursor when hovering a plot"""
     def __init__(self,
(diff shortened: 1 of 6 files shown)


pmldrmota:
Now also closes #102.

PR #371, opened 2023-10-19
Use WheelEditor from artiq.gui

ljstephenson:
see https://github.com/m-labs/artiq/commit/13271cea64302b4a3bd6e7ecef358cd52897ac61


dnadlinger:
Thanks – merged with the typo in the commit message fixed.

It's a bit annoying that this is yet another breaking change. Should be fine for us internally, though, as most users would just update the Poetry metapackage…

Issue #153, opened 2020-04-01
Running a subscan several times in run_once not supported


pmldrmota:
Suppose I want to compare on the same pair of axes the result of a subscan for different cases. I would then repeat in run_once
1. run the subscan 
2. push a result based on the scan into result channel i.
There would be one point per subscan, and as many points per run_once as the number of result channels, but ndscan doesn't know which subscan belongs to which point.

Ideally, one would have an option "Compare" in addition to "Fixed", "Refine",... that would run a point for all combinations of "Compare" parameters and plot them on the same axes with different point syles (*, o, + etc, labelled by a legend showing which combination of parameters corresponds to which style).

dnadlinger:
This definitely shouldn't be implemented as being able to call run_once() multiple times (at least, one would need a mechanism to swap between multiple sets of result channels).

More importantly, it seems like this could just be implemented on the applet side by allowing a 2D scan to be displayed as a 1D scan with two sets of series ("virtual result channels" (?)) for each value of the "comparison" scan axis. Generalises through multiple applications to n-dim. -> (n-k)-dim.

hartytp:
> More importantly, it seems like this could just be implemented on the applet side by allowing a 2D scan to be displayed as a 1D scan with two sets of series ("virtual result channels" (?)) for each value of the "comparison" scan axis. Generalises through multiple applications to n-dim. -> (n-k)-dim.

Yeah, that would be a really nice feature.

Maybe a bit orthogonal, but a nice thing to have would be the ability to fit along one axis and then plot a fit result v the other axis in a 2D scan. At present this is something one has to special case using subscans.

PR #374, opened 2023-10-22
Bool/Enum scans

pmldrmota:
Simple extension using the existing `ListGenerator`.
The same could be achieved by swapping the `BoolParam` for an `IntParam` and scanning the values `0,1`.
The plotting could be improved by labelling ticks `True`, `False`, but I expect this type of scan to be performed mainly for comparisons along the lines of #153.

### DIFF
--- a/ndscan/experiment/scan_runner.py
+++ b/ndscan/experiment/scan_runner.py
@@ -229,6 +229,9 @@ class HostScanRunner(ScanRunner):
                     for (sink, value) in zip(self._axis_sinks, axis_values):
                         # Now that we know self._fragment successfully produced a
                         # complete point, also record the axis coordinates.
+                        if isinstance(value, np.bool_):
+                            # Convert as `np.bool_` is not PYON-serializable.
+                            value = bool(value)
                         sink.push(value)
 
                     if self.scheduler.check_pause():
@@ -435,6 +438,9 @@ class KernelScanRunner(ScanRunner):
         # the axis coordinates.
         values = self._current_chunk.pop(0)
         for value, sink in zip(values, self._axis_sinks):
+            if isinstance(value, np.bool_):
+                # Convert as `np.bool_` is not PYON-serializable.
+                value = bool(value)
             sink.push(value)
 
         # Prepare for the next point.
--- a/ndscan/experiment/scan_runner.py
+++ b/ndscan/experiment/scan_runner.py
@@ -229,9 +229,6 @@ class HostScanRunner(ScanRunner):
                     for (sink, value) in zip(self._axis_sinks, axis_values):
                         # Now that we know self._fragment successfully produced a
                         # complete point, also record the axis coordinates.
-                        if isinstance(value, np.bool_):
-                            # Convert as `np.bool_` is not PYON-serializable.
-                            value = bool(value)
                         sink.push(value)
 
                     if self.scheduler.check_pause():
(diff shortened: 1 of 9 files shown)


ndscan/experiment/scan_runner.py
```diff
@@ -435,6 +438,9 @@ def _point_completed(self):
         # the axis coordinates.
         values = self._current_chunk.pop(0)
         for value, sink in zip(values, self._axis_sinks):
+            if isinstance(value, np.bool_):
+                # Convert as `np.bool_` is not PYON-serializable.
+                value = bool(value)
```
dnadlinger, line 443: What's going on here (and above)? Why are these `np.bool_`s in the first place?

pmldrmota: Probably because of `ListGenerator.points_for_level` (line 209).

pmldrmota: `LinearGenerator` returns `.tolist()` - should probably just do that in `ListGenerator`, too.

pmldrmota:
I've implemented both `bool` and `enum`-type scans, such that categorical axis labels are placed in plots.
In principle, the categorical labels could be applied to result channels as well by adding the `"categories"` key to the `spec`.
`Enum` scans don't refer to `enum.Enum` on the experiment-side yet, but I expect the `argument_editor` and `plots` implementation to still be valid even then.
At the moment, `EnumParam` behaves like `EnumerationValue` in ARTIQ, i.e. just taking on string values.

pmldrmota:
This version enables `enum.Enum` to be used for `EnumParam`s. For every `enum`, the parameter, the handle, and the parameter store need to be created. for this, a factory function is provided.

```python
@unique
class MyEnum(Enum):
    first = auto()
    second = auto()
    third = auto()

(MyEnumParam, MyEnumParamHandle, MyEnumParamStore) = enum_param_factory(MyEnum)
```

If the `Enum` values are strings, those strings will be used as axis tick labels and as options displayed in the argument interface.
In all other places, the `enum.name` is used to identify an option (in order to coerce it into a member using `MyEnum[name]`): when serialising the selection/scan from the UI and when determining the default value from a dataset entry. In the above example, the dataset default could be specified as `"dataset('baz', 'third')"` or as `"'third'"`, or as `MyEnum.third`.

Here is an example of scanning a `BoolParam` and a custom `EnumParam`,
```python
@unique
class MyEnum(Enum):
    first = "First option"
    second = "Second option"
    third = "Third option"
```
image

image

pmldrmota:
NB I haven't tested this for on-kernel scans yet.

dnadlinger:
I have a few small naming nits etc. as well, but one main issue is that we need a better solution for the type lookup. `id(…)` is not stable across time, e.g. different Python invocations such as repository scanning vs actual executation. Either this doesn't work right now, or it does, but then the `id()` output shouldn't be necessary in the first place. This may require rearchitecting ndscan's parameter store initialisation internals a bit.

I've put a rebased version here to avoid duplication of work, but didn't get to actually tackle this today: https://github.com/OxfordIonTrapGroup/ndscan/tree/wip-enum-scans

dnadlinger:
I have on-device scans working, but it requires an ARTIQ compiler fix to be able to handle scans of multiple different `EnumParam` types.

dnadlinger:
Merged in https://github.com/OxfordIonTrapGroup/ndscan/pull/399.

PR #393, opened 2024-04-12
fragment: Optimise subfragment setup/cleanup emission

dnadlinger:
All in all, this saves several seconds of compile time on the
currently biggest experiments (hundreds of fragments).

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -1,4 +1,4 @@
-from artiq.language import HasEnvironment, kernel, kernel_from_string, portable
+from artiq.language import HasEnvironment, kernel, kernel_from_string, portable, rpc
 from collections import OrderedDict
 from collections.abc import Iterable
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


hartytp:
Cool.

One comment: I find the lifecycle of `Fragment`s a little confusing in terms of "what can changes can I make safely at a given point in the experiment lifecycle"? With this change, whether or not a fragment has a trivial device setup/cleanup needs to be known at build time. That's probably fine (I can't think of any real-world use cases where I'd want to change this later on), but things like that do sometimes change at odd points, and it's not always easy to figure out whether it's safe. For example, when is it safe to modify analyses (e.g. to configure fits)? For this kind of thing, I find myself having to go and look at all the runners to figure out what they call when.

I wonder if it might make sense to introduce an explicit "finalize" step in the fragment lifecycle? That would be a natural place for this kind of code.

SquidDev:
Not sure if helpful, but somewhat related to this change — at OxIonics we recently moved all the `kernel_from_string` calls out from `build` into the `prepare` stage. This shaves a small amount off repo scan times[^1], as it avoids having to generate the kernel functions during a scan.

PR #395, opened 2024-04-19
387 clean reset of applet state when scan is overwritten

pmldrmota:
When the `source_id` changes, just reset the applet state.

### DIFF
--- a/ndscan/plots/model/subscriber.py
+++ b/ndscan/plots/model/subscriber.py
@@ -37,14 +37,14 @@ class SubscriberRoot(Root):
         if schema_revision is None:
             return
 
-        fqn = d("fragment_fqn")
-        if not self._title_set or self._context.get_title() != fqn:
+        if not self._title_set:
+            fqn = d("fragment_fqn")
             if fqn:
                 self._context.set_title(fqn)
                 self._title_set = True
 
-        source_id = d("source_id")
-        if not self._source_id_set or self._context.get_source_id() != source_id:
+        if not self._source_id_set:
+            source_id = d("source_id")
             if source_id:
                 self._context.set_source_id(source_id)
                 self._source_id_set = True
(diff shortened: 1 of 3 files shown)


pmldrmota:
Alternatively to [e6fef59](https://github.com/OxfordIonTrapGroup/ndscan/pull/395/commits/e6fef59e87a18e6ffdf6157dba87818320bb9b21), we could only reset the applet `if source_id and self._context.get_source_id() != source_id` without changing the unittest.

dnadlinger:
I wonder whether this is enough. Ideally, we'd want to clear out all of `values` at the beginning of a new scan, so we don't accidentally show old data.

pmldrmota:
It cleared it with a `SinglePointModel`, but it didn't do the trick for a `ScanModel`.

I added two commits that both individually should solve the problem (pick one?)
The first is changing the `TopLevelRunner` to clear out the point datasets at the start of a new scan.
The second is filtering the points from the `values` on the applet side.

dnadlinger:
What happens if one scan with an explicit prefix is scheduled on top of a second scan with the same explicit prefix? This is something we quite possibly can't handle cleanly, but maybe the solutions differ there.

In general, this is some tricky order-dependent code, so we should doucment the constraints assumed by the solution somewhere, at least in the form of some unit tests (like for the single-point `point_phase` mechanism, though ideally with more comments).

pmldrmota:
We could, however, catch that in `TopLevelRunner.run()` using
```python
    def run(self):
        """Run the (possibly trivial) scan."""
        previous_completed = self.get_dataset(self.dataset_prefix + "completed", default=True)
        if not previous_completed:
            raise Exception("TopLevelRunner with same explicit prefix is still running")
```
Scans usually have a new dataset where `completed` doesn't exist - hence the default - and if they do, then `completed = False`.

PR #402, opened 2024-05-02
plots.xy_1d: ignore nan datapoints when averaging

pmldrmota:


dnadlinger:
It would probably be helpful to explicitily document NaN handling somewhere, either in the `ndscan.experiment.parameters` module-level API docs, or an additional user's guide page. Something like (with better grammar): "You always need to push a value to each result channel (struct of arrays dataset storage; exception otherwise). For FloatChannels, it is perfectly acceptable to use NaN as a fallback value if there is no result to report for a certain point (e.g. because a probabilistic event did not happen); the ndscan applet will handle this correctly and omit the point from plots entirely."

PR #407, opened 2024-05-08
plots.xy_1d: adjustable num_samples_per_point

pmldrmota:
Added QSpinBox to context menu so that user can get more accurate error bars when averaging.

image


ndscan/plots/image_2d.py
```diff
@@ -130,11 +139,14 @@ def __init__(self, image_item: pyqtgraph.ImageItem,
...
         self.image_data = None
+        self.error_data = None
 
         #: Whether to average points with the same coordinates.
         self.averaging_enabled = False
+        #: Assumed number of samples per point for calculating the combined uncertainty.
+        self.num_samples_per_point = 1
         #: Keeps track of the running average and the number of samples therein.
-        self.averages_by_coords = dict[tuple[float, float], tuple[float, int]]()
+        self.averages_by_coords = dict[tuple[float, float], tuple[float, float, int]]()
```
dnadlinger, line 149: Can we make this a a frozen dataclass/… instead of a tuple? This is pretty opaque to the casual reader.

dnadlinger:
I would like to improve the user experience a bit here, in the sense of making it clear to the average physicist what is actually going on here. If I hadn't thought about this particular problem already, I wouldn't have a clue what "samples per point" actually means. Perhaps we can make the label "Assume[d] samples per point" instead, to make clear that there are no samples involved in the averaging process itself? We could also add a tooltip with a brief explanation, and should probably move the [discussion in the docstring](https://github.com/OxfordIonTrapGroup/ndscan/blob/2652a91b52b5ab344185384b197cb6cf6417719a/ndscan/plots/xy_1d.py#L33-L46) to some user-visible documentation (maybe just a new "the ndscan applet" page?).

Separately, I wonder whether it would be worth introducing a way to include enough information about the result distributions to do averaging and similar operations correctly directly with each result channel (either as channel metadata, or some kind of "rich FloatChannel").

PR #413, opened 2024-09-28
Transitive rebinding

charlesbaynham:
Adds support for transitive rebinding of parameters.

If fragment A has two children, B and C, support already existed for overriding `C.my_param` to depend on `B.my_param` then `B.my_param` to depend on `A.my_param`. 

This PR lets you do it the other way around, giving the same final result but reversing the call order so you can first bind `B.my_param` to `A.my_param`, then `C.my_param` to `B.my_param`. 

We found this need pop up when wanting to bind parameters between subfragments at the same hierarchical level.

### DIFF
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -579,21 +579,17 @@ class Fragment(HasEnvironment):
         # rebindings until a free parameter is reached.
         toplevel_source = source.owner._find_param_source(source.name)
 
-        # We don't support "transitive" binding for parameters that are already bound.
-        assert source.name in source.owner._free_params, \
-            "Source parameter is not a free parameter; already bound/overridden?"
-
         own_type = type(self._free_params[param_name])
-        source_type = type(source.owner._free_params[source.name])
+        source_type = type(toplevel_source.owner._free_params[toplevel_source.name])
         assert own_type == source_type, (
             f"Cannot bind {own_type.__name__} '{param_name}' " +
             f"to source of type {source_type.__name__}")
 
         del self._free_params[param_name]
 
-        self._rebound_own_params[param_name] = source
+        self._rebound_own_params[param_name] = toplevel_source
 
-        source.owner._rebound_subfragment_params.setdefault(source.name, []).extend(
+        toplevel_source.owner._rebound_subfragment_params.setdefault(toplevel_source.name, []).extend(
             self._get_all_handles_for_param(param_name))
 
         return param
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -541,17 +541,13 @@ class Fragment(HasEnvironment):
         """
         if param_name in self._free_params:
             return getattr(self, param_name)
+        elif param_name in self._rebound_own_params:
+            p = self._rebound_own_params[param_name]
+            return p.owner._find_param_source(p.name)
         else:
-            try:
-                rebound_param = self._rebound_own_params[param_name]
-
             # Do not support binding to parameters that are overridden
-            except KeyError:
-                raise KeyError(
-                    f"Parameter '{param_name}' is not free or rebound. Was it overridden?"
-                )
-
-            return rebound_param.owner._find_param_source(rebound_param.name)
+            raise KeyError(
+                f"Parameter '{param_name}' is not free or rebound. Was it overridden?")
 
     def bind_param(self, param_name: str, source: ParamHandle) -> Any:
         """Override the fragment parameter with the given name such that its value
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -61,8 +61,13 @@ class Fragment(HasEnvironment):
         self._free_params = OrderedDict()
 
         #: Maps own attribute name to the ParamHandles of the rebound parameters in
-        #: their original subfragment.
-        self._rebound_subfragment_params = dict()
+        #: their original subfragment, for parameters of this Fragment which are
+        #: rebind targets.
+        self._rebound_subfragment_params :dict[str, list[ParamHandle]] = dict()
+
+        #: Maps own attribute name to the Fragment and attribute name that this parameter was
+        #: rebound to, for parameters of this Fragment which have been rebound.
+        self._rebound_own_params : dict[str, tuple[Fragment, str]] = dict()
 
         #: List of (param, store) tuples of parameters set to their defaults after
         #: init_params().
@@ -547,6 +552,8 @@ class Fragment(HasEnvironment):
             f"to source of type {source_type.__name__}")
 
         del self._free_params[param_name]
+
+        self._rebound_own_params[param_name] = (source.owner, source.name)
 
         source.owner._rebound_subfragment_params.setdefault(source.name, []).extend(
             self._get_all_handles_for_param(param_name))
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -536,18 +536,18 @@ class Fragment(HasEnvironment):
         :param param_name: The name of the parameter to find the target for.
         :return: The ParamHandle for the parameter in the fragment that
             has this parameter as a free parameter.
-        :raises KeyError: If the parameter is not free or rebound (e.g. it was
+        :raises AssertError: If the parameter is not free or rebound (e.g. it was
             overridden).
         """
         if param_name in self._free_params:
             return getattr(self, param_name)
-        elif param_name in self._rebound_own_params:
-            p = self._rebound_own_params[param_name]
-            return p.owner._find_param_source(p.name)
         else:
             # Do not support binding to parameters that are overridden
-            raise KeyError(
+            assert param_name in self._rebound_own_params, (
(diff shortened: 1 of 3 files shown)


dnadlinger:
Thank you very much for the PR, and apologies for letting it languish here for so long. I completely missed that this was still unanswered all this time.

I'm not quite convinced I understand why this should be necessary. In the example of rebinding between multiple sibling fragments, doesn't the parent know which is supposed to go where?

Regardless, I'm not in principle opposed to relaxing the restrictions here. In that context, though, isn't it a bit unnatural to keep distinguishing between rebound parameters and those that are just overridden? Surely, in so far as anything matters, it's that it is no longer a free parameter of the source allow rebinding to other fragments? In other words, shouldn't this just handle already rebound parameters too?

(Also merging 10fa2858928194ca8cef1afbf45542b85f4a4195 as #417; could you just rebase this PR to drop it? (happy to fix that, but I'm assuming you are very familiar with Git))

charlesbaynham:
Hey @dnadlinger , no problem in the slightest; releasing helpful libraries publically is a good deed that is always punished by more work. 

> (Also merging https://github.com/OxfordIonTrapGroup/ndscan/commit/10fa2858928194ca8cef1afbf45542b85f4a4195 as https://github.com/OxfordIonTrapGroup/ndscan/pull/417; could you just rebase this PR to drop it? (happy to fix that, but I'm assuming you are very familiar with Git))

Done. I've just rebased to master, I assume you'll squash all these commits when merged. 

> I'm not quite convinced I understand why this should be necessary. In the example of rebinding between multiple sibling fragments, doesn't the parent know which is supposed to go where?

The context for this request is a pattern that has developed in our code. We've found ourselves using a structure where a base class implements a generic experiment and then specialises it by inheriting from mixin classes. So for example, we might have a "load a red MOT and then do something to it" experiment which has "hooks" like `do_something_to_atoms_hook()`, `imaging_hook()`, post_red_narrowband_hook()` etc, all of which are defined to be no-ops. Our mixin classes then implement these, resulting in most of our experiments looking like e.g.:

```
class ClockSpecFromXXODTFrag(
    ClockRabiSpectroscopyDipoleTrapMixin,
    DoubleTrapImagingNormalised,
    EMGain,
    FLIRBlueMOTMeasurementMixin,
    XODTSingleMolassesPlusFieldRampMixin,
    OpticalPumpingWithFieldSettingDipoleTrapMixin,
    DipoleTrapWithExperiment,
):
    """
    Clock spectroscopy from dropped XXODT

    Load into an XXODT, then use the up clock beam for spectroscopy, altering the
    (single-pass) AOM.

    Image the ground state atoms, repump and image the excited state, then image
    once more for background.
    """

    @kernel
    def before_start_hook(self):
        self.before_start_hook_clockspec()
        self.before_start_hook_xodt_molasses()
```

...with the actual behaviour implemented in the mixins. That leads us to wanting to extend existing sequences sometime but not other times, so I might have a Fragment (call it `RampFrag`) which implements a ramp in one part of the experiment and I'd like to bind some parameters of this ramp to parameters of the previous ramp. The easiest way is to have `RampFrag`s daisy-chain themselves from one to the next, but this requires me to bind parameters to a previous `RampFrag` whose params are already bound. I could instead do all the binding at once, but then it would be messier to support sequences with differing numbers of `RampFrag`s. 

That's probably an unnecessairy level of detail to motivate this PR. In general, I think it's just nice to support binding in whatever direction the user fancies without restrictions on the order in which binding methods are called. 

> isn't it a bit unnatural to keep distinguishing between rebound parameters and those that are just overridden? Surely, in so far as anything matters, it's that it is no longer a free parameter of the source allow rebinding to other fragments? In other words, shouldn't this just handle already rebound parameters too?

Hmm, yes that does seem like the most general behaviour. I made this PR as a minimal extension, but yes that probably is better. If I'm understanding the code correctly then there's a slight difference between binding to an overridden / rebound param. The behaviour in both cases should be:

**Binding to an already-bound param**

* Follow the chain to locate the free param which the target param is rebound to
* Record the new target in `_rebound_own_params`
* Remove this `ParamHandle` from `_free_params`

vs **Binding to an overridden-param**

* Follow the chain to locate the top-level which the target param is rebound to. _This is not a free param, because it has been overridden_
* Record the new target in `_rebound_own_params`
* _Set the `store` for this `ParamHandle` to match the `store` for the target, top-level `ParamHandle`
* Remove this `ParamHandle` from `_free_params`


Do you agree? If so, I can implement this. 

PS I think we should rename `_rebound_own_params` and `_rebound_subfragment_params` to `_own_params_rebound_to_others` and `_own_params_targetted_by_rebindings` or something like that. `_rebound_subfragment_params` is out of date since it doesn't necessarily point at subfragments.

dnadlinger:
Thanks for the explanation of your use case. While I think about that some more: Yes, the idea is just always to bind to the ultimate parameter. If the target already has a store (overridden), then just set the store of the rebound handle to the target's. If it doesn't yet have a store (not overridden), we can only do this later, so we need to keep track of the fact that the rebind happened to enact it down the line.

Maybe instead of storing this in `Fragment` dicts, just inlining this tree in `ParamHandle` is cleaner? Each instance would have a `ParamHandle | None` link to the parent in the tree (if any), or a `set/list/…[ParamHandle]` list of children. In fact, it seems like as soon as the parent ("self is rebound to …") link is set, the list of children ("these are rebound to self") is no longer relevant (and can be cleared), as all the children would have been (transitively) forwarded to the ultimate parent already.

dnadlinger:
Yes, one would need ot keep a list of children around.

This is strictly a refactoring opportunity, though (to avoid the naming difficulties), and can be done later down the line. For now, I'd be happy to just merge your implementation after removing the restriction to non-overridden parameters.

charlesbaynham:
Hey @dnadlinger , I was stuck at home with a fever so I found some time to do this. I've implemented the moved logic you mentioned, that resulted in much cleaner code in `Fragment`.

ndscan/experiment/parameters.py
```diff
@@ -14,7 +14,7 @@
 from artiq.language import host_only, portable, units
 from enum import Enum
 from numpy import int32
-from typing import Any, TYPE_CHECKING
+from typing import Any, TYPE_CHECKING, Optional
```
dnadlinger, line 17: `Optional` can now just be `| None` instead.

ndscan/experiment/parameters.py
```diff
@@ -275,6 +283,39 @@ def changed_after_use(self) -> bool:
...
+        """
+        Get the highest level ParamHandle in the chain of bound parameters
+
+        Walks the DAG of bound parameters to find the highest level handle.
+        That may be this ParamHandle if this parameter is not rebound.
+        """
+        if self._parent_handle is None:
+            return self
+        else:
+            return self._parent_handle._get_toplevel_handle()
```
dnadlinger, line 297: Style nitpick: In `ndscan`, early return is generally preferred (i.e. just get rid of the `else`).

ndscan/experiment/parameters.py
```diff
@@ -275,6 +283,39 @@ def changed_after_use(self) -> bool:
...
+
+    @host_only
+    def _get_all_handles_for_param(self) -> list["ParamHandle"]:
```
dnadlinger, line 300: Nitpick: The name of this seems slightly misleading, as it does not actually return all the handles for the parameter this points to. `_get_all_{bound,child}_handles()`?

ndscan/experiment/fragment.py
```diff
@@ -525,36 +521,45 @@ def bind_param(self, param_name: str, source: ParamHandle) -> Any:
...
-
-        own_type = type(self._free_params[param_name])
-        source_type = type(source.owner._free_params[source.name])
-        assert own_type == source_type, (
-            f"Cannot bind {own_type.__name__} '{param_name}' " +
-            f"to source of type {source_type.__name__}")
-
+        # To support "transitive" binding of parameters, follow the chaining of
+        # rebindings until the top-level source is found.
+        toplevel_handle = source._get_toplevel_handle()
```
dnadlinger, line 539: Do we need some check here to avoid a cycle being created (or rather, since we do get the top-level handle always, a parameter being bound to itself)? Maybe an assertion that `toplevel_handle is not param_handle` would improve the user-facing error messages here? (This + tests can wait until a follow-up PR.)

test/test_experiment_fragment.py
```diff
@@ -87,6 +146,71 @@ def build_fragment(self):
...
+
+class EnumFragString(ExpFragment):
+    def build_fragment(self):
+        self.setattr_param("value_str",
+                           EnumParam,
+                           description="Enum string",
+                           default=StrOptions.first)
+
+    def run_once(self):
+        print(self.value_str.get())
```
dnadlinger, line 169: Doesn't this just add line noise when running the tests? To check whether the rebinding works, we could e.g. just set a class member to the `get()` result.

dnadlinger:
Thanks @charlesbaynham, and I hope you are feeling better now (I was also stuck in bed the last week and a half…)!

This looks good, modulo a few nitpicks – would you mind taking care of those before I merge?

PR #417, opened 2024-11-19
fragment: Remove out-of-date rebinding implementation comment [nfc]

dnadlinger:
I believe this comment is out of date, since in e.g. test_multi_rebind
_rebound_subfragment_params does indeed end up with two ParamHandles for the "value" parameter.


dnadlinger:
This is extracted from https://github.com/OxfordIonTrapGroup/ndscan/pull/413.

Issue #316, opened 2022-12-07
[RFC] param handles having a reference to the parameter


hartytp:
A few times I've wanted to access the `Param` associated with a `ParameterHandle`. One concrete use-case for this was wanting to access the default value in a generic servo framework so I could programmatically update a linked dataset. At present `ndscan` doesn't make this easy - one needs to go via `fragment._free_params`. Is there a reason not to have `ParameterHandle`s storing a reference to the parameter?

PR #418, opened 2024-11-28
Add means of getting the parameter associated with a handle 

hartytp:
This PR adds a helper method to get the parameter associated with a given handle. Changes in the associated parameter are tracked across rebinds.

closes https://github.com/OxfordIonTrapGroup/ndscan/issues/316

### DIFF
--- a/ndscan/experiment/parameters.py
+++ b/ndscan/experiment/parameters.py
@@ -10,10 +10,13 @@
 from artiq.language import host_only, portable, units
 from enum import Enum
 from numpy import int32
-from typing import Any
+from typing import Any, TYPE_CHECKING
 from ..utils import eval_param_default, GetDataset
 
 __all__ = ["FloatParam", "IntParam", "StringParam", "BoolParam", "EnumParam"]
+
+if TYPE_CHECKING:
+    from .fragment import Fragment
 
 
 class InvalidDefaultError(ValueError):
--- a/ndscan/experiment/parameters.py
+++ b/ndscan/experiment/parameters.py
@@ -7,7 +7,6 @@
 # but to hang our heads in shame and manually instantiate the parameter handling
 # machinery for all supported value types, in particular to handle cases where e.g.
 # both an int and a float parameter is scanned at the same time.
-
 from artiq.language import host_only, portable, units
 from enum import Enum
 from numpy import int32
@@ -213,12 +212,15 @@ class ParamHandle:
     :param owner: The owning fragment.
     :param name: The name of the attribute in the owning fragment bound to this
         object.
+    :param parameter: The parameter associated with this handle. The
+        :attr:`~.ParamHandle.parameter` attribute points to the parameter currently
+        associated with this handle, tracking binding of the parameter.
     """
-    def __init__(self, owner: Any, name: str):
-        # `owner` will typically be a Fragment instance; no type hint to avoid circular
-        # dependency.
+    def __init__(self, owner: "Fragment", name: str, parameter):
         self.owner = owner
         self.name = name
+        self.parameter = parameter
+
         assert name.isidentifier(), ("ParamHandle name should be the identifier it is "
                                      "referred to as in the owning fragment.")
 
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -370,7 +370,7 @@ class Fragment(HasEnvironment):
         param = param_class(fqn, description, *args, **kwargs)
         self._free_params[name] = param
 
-        handle = param.HandleType(self, name)
+        handle = param.HandleType(self, name, param)
         setattr(self, name, handle)
         return handle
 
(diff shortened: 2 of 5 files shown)


ndscan/experiment/parameters.py
```diff
@@ -213,18 +215,23 @@ class ParamHandle:
...
         assert name.isidentifier(), ("ParamHandle name should be the identifier it is "
                                      "referred to as in the owning fragment.")
 
         self._store = None
         self._changed_after_use = True
 
+    def get_parameter(self) -> "ParamBase":
+        """Returns the parameter associated with this handle."""
+        return self._parameter
+
```
dnadlinger, line 234: Can just be a public field like `owner`/`name`. Python supports properties anyway if this ever needs to change (or we really wanted to enforce read-only-ness).

dnadlinger:
Regarding tracking rebinds, I guess you have some intuition about why this is the "correct" behaviour? Doesn't seem obvious to me either way.

dnadlinger:
I'm happy to merge the unrelated small chagnes as part of this PR for simplicity, but could you rebase to merge the two first commits (to keep the unit tests passing on all commits)?

hartytp:
> Regarding tracking rebinds, I guess you have some intuition about why this is the "correct" behaviour? Doesn't seem obvious to me (but neither would the converse).

Good question.

My assumption was "If I ask a parameter handle what metadata is associated with it the answer I get should reflect what is actually going to happen when I run the experiment, rather than how the fragment was originally created." and "the metadata I get should not depend on which handle I ask".

Maybe there is a different intuition I'm missing, but those just seemed to be reasonable assumptions. Either way, I think this needs calling out explicitly in the documentation so once we've concluded this I'll update that.

> This might also limit the usefulness of this in general/require careful documentation, as the semantics may not be that obvious, and accessing the parameter metadata is almost never necessary in user code. Omitting this was not a principled decision either, but just made to keep the API surface minimal initially (in terms of pointing users in the right direction).

I've found a number of times when accessing parameter metadata has been useful in user code. For example, suppose I want to make a small framework which automatically pushes analysis results to the dataset linked to a parameter.

ndscan/experiment/parameters.py
hartytp: Happy to do that if you prefer

hartytp:
> I'm happy to merge the unrelated small chagnes as part of this PR for simplicity, but could you rebase to merge the two first commits (to keep the unit tests passing on all commits)?

Sorry, I'm not sure I follow the request here. Do you mean essentially rewriting history so I fix the tests before adding the new feature?

dnadlinger:
> Sorry, I'm not sure I follow the request here. Do you mean essentially rewriting history so I fix the tests before adding the new feature?

Just `rebase -i` to merge 27ebbd2 and 42c7372, as either alone would lead to the CI breaking. (Not that we need to bisect the history often for ndscan to track down some bugs, but gratuitous test failures can be annoying in that case.)

dnadlinger:
Great, thanks. Could you just remove the `get_parameter` accessor as well, and add one very basic assertion to the parameter tests checking that `parameter` is set to the value passed to the constructor, so we have a regression test for the existence of the API?

Happy to merge this after that and a short docstring for `ParamHandle.parameter`.

hartytp:
Done - will merge once CI passes if you're happy. Thanks for the quick review!

dnadlinger:
Huh, I was trying to check whether the `.parameters` attribute shows up as being documented (you might need to add a `#: foo` docstring before the `self.parameter = …` line), but it seems like all the `parameters` classes are missing from the API docs.

Could you rebase again please to merge the fixups into the first commit (and only leave the unrelated changes separately)?

dnadlinger:
Oh, it's probably the __all__…

dnadlinger:
Ah, the branch is on the OITG fork anyway, so I'll just fix there myself.

dnadlinger:
Thanks!

hartytp:
@dnadlinger not sure if useful, but I've found the workflows at `ionics_fits` for docs relatively useful - e.g. having a doc build as a test helps to keep the quality up (also a `poe` task is quite useful when checking things locally). I'd be happy to submit a PR if you ever wanted anything along those lines added

hartytp:
>Doc builds are already run as part of CI (and uploaded to https://oxfordiontrapgroup.github.io/ndscan/apidocs.html#module-ndscan.experiment.parameters for master commits).

aah, cool, I must have missed how this works. Does the current workflow prevent merging PRs with doc build errors (e.g. missing references)?

dnadlinger:
The doc builds are marked as "requried", so if `make html` outright fails, it will indeed block merges. Sphinx warnings probably don't currently count as errors right now.

dnadlinger:
Added a `docs` target in https://github.com/OxfordIonTrapGroup/ndscan/commit/6a229ee664a029d1b61d9079f1a5afe06f8175f8.

Issue #415, opened 2024-11-04
[RFC] Means of accessing child fragment results within a kernel


hartytp:
A use-case I've come across a number of times now is wanting to access the results of a fragment from within a kernel.

This often comes up in the context of `AggregateExpFragment` where one aggregates multiple `ExpFragment`s into a parent `ExpFragment` (with the notion of "a fragment is essentially a function" this is basically saying that when one has functions which call other functions, it's often useful to be able to act on their return values). One example of this is "I have a sideband thermometry `AggregateExpFragment` which calls a red and blue sideband and now I want to calculate nbar based on the measured populations. Another example might be I want to run a servo with a wide capture range and then use a result to seed a servo with an narrow capture range (forward a result channel to a parameter).

Since result channel sinks [should never be accessed by kernel code](https://github.com/OxfordIonTrapGroup/ndscan/issues/414), there is currently no sanctioned way of achieving this within ndscan. This thread is a request for comments to discuss a proposed implementation. I think a good implementation of this should:

1. Be ergonomic and user-friendly
2. Work naturally with the current limitations of ARTIQ python
3. Not rely on synchronous RPCs - getting results should be supported without having to leave the kernel

@dnadlinger LMK if you would be happy merging a PR along the lines of the below (and feel free to suggest names / API) and I'd be happy to write one.

# Proposal 1: Introduce an API for absorbing result channels

Absorbing result channels - taking the result channel out of the scan schema seen by the `ScanRunner` and then re-exporting - is already used by subscans. We've discussed having a convenient API for this. I think it would look like the following:

1 - We introduce a new `AbsorbedResultSink`.

```python
class AbsorbedResultSink(ResultSink):
    """ Sink used when result channels are "absorbed" by another fragment.

    This sink acts much like a :class:`LastValueSink` with the addition of forwarding all
    pushed values to the specified sink.
    """
    def __init__(self, channel: ResultChannel):
        self.channel = channel
        self.value = None

    def push(self, value: Any) -> None:
        self.channel.push(value)
        self.value = value

    def get_last(self) -> Any:
        """Return the last-pushed value, or ``None`` if none yet."""
        return self.value
```

2 - We introduce a new method `Fragment.absorb_result(self, channel: ResultChannel) -> AbsorbedResultSink`

This boils down to essentially:

```python

def Fragment.absorb_result(self, name: str, channel: ResultChannel) -> AbsorbedResultSink:
    self.setattr_result_like(name, channel)
[...]

PR #421, opened 2024-11-29
Add NumericChannel.get_last method

hartytp:
This PR introduces a new `NumericChannel.get_last` method (mirroring the interface provided by sinks).

Prior to this PR there was no means of getting the results from a fragment within a kernel (the closest option was to use `channel._sink.get_last` and hope that's supported by the sink type). This feature is useful, for example, in `ExpFragment`s which aggregate multiple other `ExpFragment`s into a single experiment and want to have some form of feedforward based on result values.

Closes https://github.com/OxfordIonTrapGroup/ndscan/issues/415

### DIFF
--- a/ndscan/experiment/result_channels.py
+++ b/ndscan/experiment/result_channels.py
@@ -291,6 +291,7 @@ class NumericChannel(ResultChannel):
     def push(self, raw_value) -> None:
         """
         """
+        self._value_pushed = True
         self._last_value = raw_value
         self._push(raw_value)
 
--- a/ndscan/experiment/result_channels.py
+++ b/ndscan/experiment/result_channels.py
@@ -2,7 +2,7 @@
 Result handling building blocks.
 """
 
-from artiq.language import HasEnvironment, portable, rpc
+from artiq.language import HasEnvironment, kernel, portable, rpc
 import artiq.language.units
 from typing import Any
 from .utils import dump_json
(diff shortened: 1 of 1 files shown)


dnadlinger:
This should be kernel-only and documented as a hacky workaround for a compiler limitation. Host experiments should still go via sinks (and we should add absorbing functionality/… as required to make that work.)

hartytp:
Done

hartytp:
>This should be kernel-only

@dnadlinger I'll leave it to you to make the final call on this, but FWIW I think it's a shame to make this kernel-only. My argument would be that I think it's really valuable to have code which is portable between the host and kernel. Keeping this kernel-only breaks that since the host version has to use an interface based on absorbing results while the kernels does this.

hartytp:
@dnadlinger gentle nudge: what do you want to see to merge this? How do you feel about making the method portable?

dnadlinger:
I'm fine with merging this now if this solves your issues, but I'm really not happy with the design still. For instance, it doesn't play well with the "muting" interface (which fragments can use to avoid creating expensive data that is just discarded). As for host-side access, note that as written this doesn't really work in a lot of use cases involving on-kernel scans, as it would require attribute writeback to take place first.

Personally, I'm still not sure whether just writing the result to a member variable to be accessed by parent fragments isn't the cleaner design. If it needs to be a hack, might as well make it obvious what's going on and avoid cluttering the API. At some point, we can just fix the compiler to properly support function pointers/delegates, and have an array of sink callbacks on the core device as well, properly fixing the issue.

hartytp:
Thanks @dnadlinger .

> Personally, I'm still not sure whether just writing the result to a member variable to be accessed by parent fragments isn't the cleaner design. If it needs to be a hack, might as well make it obvious what's going on and avoid cluttering the API.

How would this work?

> At some point, we can just fix the compiler to properly support function pointers/delegates, and have an array of sink callbacks on the core device as well, properly fixing the issue.

Yes. I completely agree that the real issue here is the ARTIQ compiler and everything else is just a hack to get around it.

Issue #416, opened 2024-11-04
[RFC] AggregateExpFragment accepting callables as well


hartytp:
I've found myself wanting a few times to add function code to `AggregateExpFragment`s as well as fragments. The uses cases here are the same kind of things discussed in https://github.com/OxfordIonTrapGroup/ndscan/issues/415

The implementation I have in mind would look like

```python
class AggregateExpFragment(ExpFragment):
    r"""Combines multiple :class:`ExpFragment`\ s and callables by executing them one
    after each other each time :meth:`run_once` is called.

    To use, derive from the class and override ``build_fragment()`` to pass in the objects
    to be aggregated::

        class FooBarBazAggregate(AggregateExpFragment):
            def build_fragment(self):
                self.setattr_fragment("foo", FooFragment)
                self.setattr_fragment("bar", BarFragment)

                # Any number of customisations can be made as usual,
                # e.g. to provide a convenient parameter to scan the
                # fragments in lockstep:
                self.setattr_param_rebind("freq", self.foo)
                self.bar.bind_param("freq", self.freq)

                @kernel
                def baz() -> None:
                    pass  # do something!

                # Let AggregateExpFragment default implementations
                # take care of the rest, e.g. have self.run_once()
                # call self.foo.run_once(), then self.bar.run_once() then baz().
                super().build_fragment([self.foo, self.bar, baz])

        ScanFooBarBazAggregate = make_fragment_scan_exp(FooBarBazAggregate)


    Each aggregated experiment should be independent, i.e. do its own setup and clean
    up.
    """
    def build_fragment(self, operands: List[Union[Callable[[], None],
                                                  ExpFragment]]) -> None:
        """
        :param operands: The list of objects to be aggregated. Each item may either be
            a "child" fragment or a callable. The operands will be run in the given
            order, calling ``run_once`` on child fragments. No special treatment is
            given to the ``{host,device}_{setup,cleanup}()`` methods, which will be
            executed through the recursive default implementations unless overridden by
            the user.
        """
```

@dnadlinger LMK if you have comments or suggestions. If you'd merge a PR along these lines I'd be happy to submit one.

dnadlinger:
Yep something like this looks good to me.

The unit tests should exercise mixed @kernel and host Python callables just to cover the API surface.

hartytp:
Thanks. Agreed re tests.

[...]

PR #425, opened 2024-12-09
Let AggregateExpFragments take callables as well as exp fragments

hartytp:
Closes https://github.com/OxfordIonTrapGroup/ndscan/issues/416

See #416 for discussion

I haven't added a kernel test for this but would be happy to do so if requested. Not sure the best way of doing this - perhaps making all of `AddOneFragment`'s methods protable and then having `AddOneAggregate` take a build-time boolean flag to determine whether it runs on the kernel or host? Suggestions welcome!

### DIFF
--- a/test/fixtures.py
+++ b/test/fixtures.py
@@ -202,7 +202,13 @@ class AddOneAggregate(AggregateExpFragment):
     def build_fragment(self) -> None:
         self.setattr_fragment("a", AddOneFragment)
         self.setattr_fragment("b", AddOneFragment)
-        return super().build_fragment([self.a, self.b])
+
+        self.setattr_result("sum", FloatChannel)
+
+        def push_sum():
+            self.sum.push(self.a.value.get() + 1 + self.b.value.get() + 1)
+
+        return super().build_fragment([self.a, self.b, push_sum])
 
 
 class TrivialKernelAggregate(AggregateExpFragment):
--- a/ndscan/experiment/fragment.py
+++ b/ndscan/experiment/fragment.py
@@ -3,7 +3,7 @@ from collections import OrderedDict
 from collections.abc import Iterable
 from copy import deepcopy
 import logging
-from typing import Any
+from typing import Any, Callable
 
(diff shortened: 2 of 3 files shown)


test/fixtures.py
```diff
@@ -202,7 +202,13 @@ class AddOneAggregate(AggregateExpFragment):
     def build_fragment(self) -> None:
         self.setattr_fragment("a", AddOneFragment)
         self.setattr_fragment("b", AddOneFragment)
-        return super().build_fragment([self.a, self.b])
+
+        self.setattr_result("sum", FloatChannel)
+
+        def push_sum():
+            self.sum.push(self.a.value.get() + 1 + self.b.value.get() + 1)
```
hartytp, line 209: I would love this to be `self.sum.push(self.a.result.get_last() + self.b.result.get_last())` but:
* currently the `get_last` is deliberately kernel only
* we could add a sink-based interface that works by absorbing results...but then that's host-only so we can't run the test in a portable fashion
* I also considered having `AddOneFragment` store a `last_value` attribute and store the result there as well, but that also seems like a pretty miserable hack. We're essentially saying to users "getting the last value pushed to a result is a legitimate thing to do, but we won't support it so you have to manually duplicate result channels and handle all the accompanying synchronization issues yourself"

dnadlinger: > We're essentially saying to users "getting the last value pushed to a result is a legitimate thing to do, but we won't support it so you have to manually duplicate result channels and handle all the accompanying synchronization issues yourself"

This is still preferrable to adding a half-baked implementation with unclear semantics to ndscan. "Magic" solutions that work only most of the time with unclear semantics (i.e. leaky abstractions) are much worse than just requiring users to write some straightforward Python code.

dnadlinger:
Thanks, merged with a bug in the docstring (override_param() return value order) fixed.

As for a test case, I'd say the dozen lines it takes to just add a separate `@kernel` copy of the fragment is probably worth it over the added complexity of some kind of parameterised solution.

PR #430, opened 2025-01-28
Copy apply_color method from ARTIQ _ArgumentEditor to ndscan's

william-cutler:
ArgumentEditor to ensure compatibility with this ARTIQ change on October 8, 2024:

https://github.com/m-labs/artiq/commit/52c07a2b145b436053f8072eca7300759b690801


dnadlinger:
Thanks a lot for tracking this down yourself – I really appreciate help maintaning the library!

I've added a second commit to add the other relevant part of the upstream change (but only if available, to keep backwards-compatibility).

PR #444, opened 2025-03-09
Add option to repeat each point (before changing params) in a scan

thatschatt:
This adds a scan option to repeat each point `n` times before moving on to the next point. Repeating the whole scan is not affected (you can repeat each point `n` time, and also repeat the whole scan `m` times).

This is useful for running scans where there is some settling time every time parameters are changed, without having to resort putting your own loops inside the fragments.

### DIFF
--- a/ndscan/results/arguments.py
+++ b/ndscan/results/arguments.py
@@ -83,7 +83,8 @@ def dump_scan(schema: dict[str, Any]) -> Iterable[str]:
         path = ax["path"] or "*"
         yield f"   - {ps['description']} ({fqn}@{path}):"
         yield f"     {format_scan_range(ax['type'], ax['range'], ps)}"
-    yield f" - Number of repeats: {scan['num_repeats']}"
+    yield f" - Number of repeats of scan: {scan['num_repeats']}"
+    yield f" - Number of consecutive repeats of each point: {scan['repeat_each_point']}"
     yield f" - Randomise order globally: {scan['randomise_order_globally']}"
 
 
--- a/ndscan/results/arguments.py
+++ b/ndscan/results/arguments.py
@@ -84,7 +84,7 @@ def dump_scan(schema: dict[str, Any]) -> Iterable[str]:
         yield f"   - {ps['description']} ({fqn}@{path}):"
         yield f"     {format_scan_range(ax['type'], ax['range'], ps)}"
     yield f" - Number of repeats of scan: {scan['num_repeats']}"
-    yield f" - Number of consecutive repeats of each point: {scan['repeat_each_point']}"
+    yield f" - Number of repeats per point: {scan['num_repeats_per_point']}"
     yield f" - Randomise order globally: {scan['randomise_order_globally']}"
 
 
--- a/ndscan/dashboard/argument_editor.py
+++ b/ndscan/dashboard/argument_editor.py
@@ -74,24 +74,24 @@ class ScanOptions:
 
         #
 
-        self.repeat_each_point_container = QtWidgets.QWidget()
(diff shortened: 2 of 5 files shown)


dnadlinger:
Hi @thatschatt, many thanks for the PR! This isn't something we would use very frequently here, but on balance, I can see it being useful often enough to be worth going into ndscan itself.

Could you just update the changes to use the same naming conventions as the existing code, please? It's perhaps not very well-documented, but `ndscan` uses `num_<plural>` for numbers of things throughout. Maybe something like `num_repeats_each_point`, `num_repeats_per_point`, `num_per_point_repeats`, …

Unrelatedly, I'm just curious: Which experiment/… are you using ndscan for?

thatschatt:
`num_repeats_per_point` works for me.

We're using NDScan for an update/rewrite of the code running our Ba+ ion trap here in Aarhus. Our software there has so far taken the predictable course of slowly morphing some some simple 'test' scripts into a huge monolith that does everything. And I am very much dreading the day the student who made that monolith graduates. So we've picked NDScan as the basis for the new system, on the grounds that looks like someone smart has already thought about a lot of the problems that come up, so this saves us a lot of work :)

Currently we're just working on getting all the basics up and running, e.g. Rabi flopping and spectroscopy scans. It was here that the option to repeat points comes in real handy, because this way we can implement sequences without any internal loops in the code. 

In the coming weeks, I may well send you a lot of questions about good practices/how you guys implement NDScan workflows, if that's ok? 

- Adam

dnadlinger:
Great, thanks – I think there is just one left-over from refactoring, but then I'm happy to merge. One thing that would be great to have are some simple regression tests in `test/` to make sure this behaves as intended. `test/test_experiment_scan_generator.py` would be the natural place, but it seems so far I've neglected to add a unit test for `generate_points` on top of the scan generators. Would you have time to add that – either here or in a follow-up PR? Just a quick check that the combinations of 1 and 2 `num_repeats` and `num_repeats_per_point` work as intended would already be enough to catch catastrophic breakage.

And yes, of course I'm happy to answer any questions regarding ndscan as you start rolling it out in the lab – feel free to contact me, or use the new [Discussions](https://github.com/OxfordIonTrapGroup/ndscan/discussions) tab. Enabling a maintainable codebase (we have ~100k LOC on the networking experiment here) was indeed one of the motivations behind ndscan, so I hope it can be helpful for you too. As for multiple runs of the same point to get the readout statistics (we call them "shots" here), the reason why we don't mind having an explicit loop (abstracted in an ExpFragment) for those is for run-time optimisation purposes; often e.g. DMA sequences do not need to be re-recorded, devices configured, etc. to just run the same sequence 50 or 500 times.

ndscan/results/arguments.py
```diff
@@ -83,7 +83,8 @@ def dump_scan(schema: dict[str, Any]) -> Iterable[str]:
         path = ax["path"] or "*"
         yield f"   - {ps['description']} ({fqn}@{path}):"
         yield f"     {format_scan_range(ax['type'], ax['range'], ps)}"
-    yield f" - Number of repeats: {scan['num_repeats']}"
+    yield f" - Number of repeats of scan: {scan['num_repeats']}"
+    yield f" - Number of repeats per point: {scan['num_repeats_per_point']}"
```
dnadlinger, line 87: @thatschatt: Just for future reference (I didn't notice this in the original review), this shouldn't have been expecting `num_repeats_per_point` to exist to stay backwards-compatible with older results files. Fix at https://github.com/OxfordIonTrapGroup/ndscan/pull/456.

Issue #3, opened 2019-02-08
Save state of scan controls in argument UI


dnadlinger:
Currently, all parameters are set to `Fixed` and scan settings cleared when an experiment is reopened in the dashboard (or the dashboard is restarted).

Saving/restoring the state of the scan type controls would considerably improve quality of life.

PR #448, opened 2025-05-20
Add scan parameter loading into UI

thatschatt:
This PR restores the state of the scan parameters into the the arguments window when the window is reopened or when an H5 results file is loaded.

Because there was no distinction between a min/max scan or a centered scan in the results, I have added a `style` attribute for the range scans so that the correct type can be reloaded. This defaults to min/max when loading results files with no style tag.

I've tested it with loading from an H5 file for all scan types:
- min/max linear
- min/max refining
- centre linear
- centre refining
- expanding
- list
- bool
- enum

Also tested are multidimensional scans in a few combinations.

This closes issue #3


dnadlinger:
Thanks, this should be a great usability improvement! Somehow, we never seem to load in old files here as part of our workflow, so this never quite made it to the top of my priority list, but even for us, the retention of settings across closing/reopening the experiment or dashboard will be nice.

Implementation-wise, I'm not so sure about the second enumeration of all the types in `get_scan_option_class` and `isinstance` checks. Not just the duplication itself, but in particular the bool/enum gymnastics seem a bit brittle. Could we instead make `read_from_axis(dict) -> None` a `try_read_from_axis(dict) -> bool`, and just iterate through all the registered options until we find one that matches (return `True`)?

dnadlinger:
(`read_matching_axis(dict) -> bool`, maybe)

thatschatt:
I agree, the separate function to work out the class was a little clunky. I've now swapped it to `attempt_read_from_axis(dict) -> bool`, implemented for all the options that could scan.

thatschatt:
(for future reference: is it polite / good practice to rebase PRs onto the main branch while working on them?)

dnadlinger:
I'm not sure I understand where the problem is here. A list is, for any data type, just a list of values. We already know what the type is, because we are now matching against the existing options, which will have been populated according to the parameter definition.

(The only potential issue would be if the type of a parameter was changed, but it's completely fine to fail to load then, in which case ARTIQ I think will throw away the client-side state in the same way as it would on Recompute All Arguments.)

thatschatt:
sounds good to me, thanks!

ndscan/dashboard/argument_editor.py
```diff
@@ -1134,10 +1239,8 @@ def write_to_params(self, params: dict) -> None:
                 "randomise_order": self.check_randomise.isChecked()
             }
         }
-        if (lower := schema["spec"].get("min", None)) is not None:
-            spec["range"]["limit_lower"] = lower
-        if (upper := schema["spec"].get("max", None)) is not None:
-            spec["range"]["limit_upper"] = upper
+        spec["range"]["limit_lower"] = self.min / self.scale
+        spec["range"]["limit_upper"] = self.max / self.scale
```
dnadlinger, line 1243: Dividing by scale here looks wrong (this is following the ARTIQ convention where values are in base units and scale is just applied for display purposes).

dnadlinger:
Force-pushed some updates to your branch (fixed that scale issue pointed out in the comment earlier, removed some code duplication), but need to test before merging still (-> tomorrow).

thatschatt:
Looks good to me, thanks! Not sure how the scale issue slipped through there, I was pretty sure I tested rather thoroughly...

The only remaining issue (if it is an issue) is that if you load an h5 file from an experiment with different parameters, the window updates to reflect the state it was in when that experiment was in back then, and hides the new parameters. If a parameter was deleted then we see an error on submission, but if new parameters were added to the experiment then it just silently picks up the default. I guess one solution here would be to run `recompute_arguments` before we fill in the new scan items, and then check for parameters which appear in one list but no the other. This doesn't feel too clean to me, but I can't immediately think of a nice way to include the param as a default or at least warn the user,

dnadlinger:
Thanks for having another look; testing this again was on my list for later today.

Yes, this is an artefact of ndscan encoding all the parameter definitions in the default value of the `PARAMS_ARG_KEY` parameter. We of course already have our custom argument editor code, so changing the behaviour would be doable in principle, but as we currently just use the experiment dock state handling functions, it makes sense that this would be the result.

Since I basically never use the HDF5 file loading function, I'm not actually sure how this is handled for upstream ARTIQ – IIRC the arguments are just passed as "overrides" to the recompute arguments code?

I suppose we could connect the button to our own implementation instead that removes `PARAMS_ARG_KEY` before passing it on to the dock's recompute arguments implementation (and add matching overrides/scan settings afterwards.

PR #458, opened 2025-08-03
ndscan.show: Display experiment class name in metadata

ayushagrawal2:


dnadlinger:
Thanks! Could you rename `AritqMetadata.exp_class` to `class_name` to follow the rule of least surprise – all the other members match the name in the source (though granted this is nested inside `expid`). Otherwise LGTM.

ayushagrawal2:
Sure, done!

PR #469, opened 2026-01-16
Support for saving OnlineFit results

ayushagrawal2:
experiment.default_analysis: Add support for automatically pushing OnlineFit results to analysis_results

Default behaviour:
* If `oitg` is [#a24672](https://github.com/OxfordIonTrapGroup/oitg/commit/a2467207768ec7bed1a71c3a8fbba91eee407332) or newer:
    * Saves all params
* Else:
    * If explicitly requested to save true:
         * Saves non-derived params
    * No save mode specified:
         * Does not save any params

### DIFF
--- a/ndscan/experiment/default_analysis.py
+++ b/ndscan/experiment/default_analysis.py
@@ -21,6 +21,9 @@ import logging
 from collections.abc import Callable, Iterable
 from typing import Any
 
+import numpy as np
+from oitg.fitting.FitBase import FitBase, FitError
+
 from ..utils import FIT_OBJECTS
 from .annotations import (
     Annotation,
@@ -30,7 +33,7 @@ from .annotations import (
     computed_curve,
 )
 from .parameters import ParamHandle
-from .result_channels import ResultChannel
+from .result_channels import OpaqueChannel, ResultChannel
 
 __all__ = [
     "Annotation",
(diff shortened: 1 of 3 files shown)


ndscan/experiment/default_analysis.py
```diff
@@ -327,7 +366,42 @@ def execute(
...
+
+        try:
+            p_dict, p_error_dict, residuals = fitter.fit(
+                x, y, y_err, calculate_residuals=True
+            )
+        except FitError:
+            logger.warning("Fit failed for fit type '%s'", self.fit_type, exc_info=True)
+            return []
+
+        reduced_chi_squared = np.sum(residuals**2) / (len(y) - len(p_dict))
```
dnadlinger: This definition seems a bit odd/not helpful in general without further clarification – wouldn't the squared residuals usually be scaled by the estimated variance (y_err**2) to yield "the chi-squared"? We might not always have y_err, of course, but when we do, doesn't the scaled one seem more appropriate for the usual goodness-of-fit measure? (Ideal around 1 for good fit with properly estimated input errors.)

ndscan/experiment/default_analysis.py
```diff
@@ -241,10 +245,46 @@ def __init__(
...
         initial_values: dict[str, Any] | None = None,
+        save_fit_results: bool | None = None,
     ):
         self.fit_type = fit_type
+
+        params = []
         if fit_type not in FIT_OBJECTS:
-            logger.warning("Unknown fit type: '%s'", fit_type, exc_info=True)
+            raise ValueError(f"Unrecognised fit type: '{fit_type}'")
+        else:
```
dnadlinger: Can now remove this else block.

ndscan/experiment/default_analysis.py
```diff
@@ -327,7 +366,42 @@ def execute(
...
+        x = _extract_data(self.data["x"])
+        y = _extract_data(self.data["y"])
+        y_err = _extract_data(self.data.get("y_err", None))
+
+        fitter: FitBase = FIT_OBJECTS[self.fit_type]
+
+        try:
+            p_dict, p_error_dict, residuals = fitter.fit(
+                x, y, y_err, calculate_residuals=True
+            )
```
dnadlinger: Seems like this is missing `constants`/`initial_values`.

ndscan/experiment/default_analysis.py
```diff
@@ -327,7 +369,53 @@ def execute(
...
+                calculate_residuals=True,
+                constants=self.constants,
+                initialise=self.initial_values,
+            )
+        except FitError:
+            logger.warning("Fit failed for fit type '%s'", self.fit_type, exc_info=True)
+            return []
+
+        weights = 1 / (np.asarray(y_err) ** 2) if y_err is not None else np.ones_like(y)
+        reduced_chi_squared = np.sum(residuals**2 * weights) / (len(y) - len(p_dict))
```
dnadlinger: Should this be only the `len()` of the free parameters, not of the entire `p_dict`?

ayushagrawal2: ah yes, also need to normalise the weights

test/test_experiment_default_analysis.py
```diff
@@ -20,3 +52,246 @@ def make_axes(*axes):
...
+                continue
+
+            logger.warning(
+                f"param_name={param_name}, value={value}, actual_params={actual_params[param_name]}"
+            )
+            self.assertTrue(
+                relative_error(actual_params[param_name], value) <= TOLERATED_FIT_ERROR,
+                f"Parameter '{param_name}' fit value {value} deviates from actual "
+                f"value {actual_params[param_name]} beyond tolerated error. {result.sink.get_last()}",
+            )
```
ayushagrawal2: Can't seem to get these tests to work. I tried using an `ArraySink` in `_create_online_fit` below, and it looks like all results pushed to the same channel, and I couldn't recover the order of which index corresponds to which param. Dumping out the array and manually inspecting shows that the values all match.

This is of course not an issue in the actual implementation. Any pointers on how to retrieve the analysis result values in the test?

ndscan/experiment/default_analysis.py
```diff
@@ -327,7 +369,57 @@ def execute(
...
+        )
+
+        for param, value in p_dict.items():
+            param_name = self._channel_prefix + param
+            if param_name not in self._result_channels:
+                continue
+            self._result_channels[param_name].push(value)
+            self._result_channels[param_name + "_err"].push(
+                p_error_dict.get(param, None)
+            )
```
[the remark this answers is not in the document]
ayushagrawal2: Fixed: Just omitting missing error channels entirely

PR #493, opened 2026-04-22
argument editor: Add support for tooltips to param labels

ayushagrawal2:
Thought it would be nice to allow more verbose descriptions in a tooltip on parameter labels like vanilla artiq

### DIFF
--- a/ndscan/dashboard/argument_editor.py
+++ b/ndscan/dashboard/argument_editor.py
@@ -453,7 +453,7 @@ class ArgumentEditor(QtWidgets.QTreeWidget, OverrideProvider):
         if explanation := schema.get("explanation", ""):
             label.setToolTip(explanation)
         else:
-            label.setToolTip(schema["description"])
+            label.setToolTip(schema["description"] + " (no <i>explanation</i> set)")
 
         # For whatever reason, the auto-sized column is not wide enough to display the
         # whole label if displayed through a widget – whether through an extra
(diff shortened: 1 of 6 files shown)


ndscan/experiment/parameters.py
```diff
@@ -420,6 +422,7 @@ def describe(self) -> dict[str, Any]:
             "type": "float",
             "default": str(self.default),
             "spec": spec,
+            "tooltip": self.tooltip,
```
ayushagrawal2: Worth refactoring so the fqn, description, and tooltip are handled by the base class in its own describe method?

ndscan/dashboard/argument_editor.py
```diff
@@ -450,6 +417,11 @@ def add_item(widget_item):
         label.setFont(font)
         label_container.addWidget(label)
 
+        if schema.get("tooltip", ""):
+            label.setToolTip(schema["tooltip"])
```
dnadlinger:
```
if (tooltip := schema.get("tooltip")):
  label.setToolTip(tooltip)
```

ayushagrawal2:
Re: renaming `tooltip` -> `help` as discussed offline, python already has a built-in function `help()`. Maybe `verbose_description` is better?

dnadlinger:
Hmm… `help_text`? `long_description`? `extended_description`? `guidance`? `explanation`? `comment`?

ayushagrawal2:
`help_text` it is

dnadlinger:
Apologies for the bike-shedding, but I decided that I liked "explanation" better in context (e.g. it isn't "description_text" either, and going with the slightly opinionated nature of ndscan, will perhaps nudge people towards the intended use). I've also merged the commits into one as I prefer keeping associated unit tests in the same commit (when merging, etc.).

I've also added a callout as to how the tooltip can be set, and added the explanations to the param tree dialog (if present).

Issue #11, opened 2019-02-08
Support scans for integer parameters


dnadlinger:
Currently, all scans calculations are done using floating point – need to appropriately round step size, and stop refining scans once all integers have been covered.

dnadlinger:
See also https://github.com/OxfordIonTrapGroup/ndscan/issues/154.

PR #494, opened 2026-04-23
Support scans for integer parameters

ayushagrawal2:
Implements #11

### DIFF
--- a/ndscan/experiment/scan_generator.py
+++ b/ndscan/experiment/scan_generator.py
@@ -3,6 +3,7 @@ from collections.abc import Iterator
 from dataclasses import dataclass
 from itertools import product
 from typing import Any
+from venv import logger
 
 import numpy as np
 
@@ -12,7 +13,13 @@ __all__ = [
     "LinearGenerator",
     "ListGenerator",
     "CentreSpanRefiningGenerator",
+    "IntRefiningGenerator",
+    "IntLinearGenerator",
+    "IntCentreSpanGenerator",
+    "IntCentreSpanRefiningGenerator",
     "ScanOptions",
+    "GENERATORS",
+    "INT_GENERATORS",
 ]
 
 
@@ -69,6 +76,21 @@ class RefiningGenerator(ScanGenerator):
         target["min"] = self.lower
         target["max"] = self.upper
 
+    @staticmethod
+    def args_from_centre_span(
+        centre, half_span, randomise_order, limit_lower, limit_upper
+    ):
+        lower = centre - half_span
+        upper = centre + half_span
+        if lower < limit_lower and upper > limit_upper:
+            # If both extents of the span exceed the limits, change the span to fit
+            #   the larger of the two. This way we still don't move the scan centre.
+            half_span = max(abs(centre - limit_lower), abs(limit_upper - centre))
+            lower = centre - half_span
+            upper = centre + half_span
+
+        return lower, upper, randomise_order
+
 
 class CentreSpanRefiningGenerator(RefiningGenerator):
     """Generates progressively finer grid in span around centre by halving distance
@@ -88,19 +110,14 @@ class CentreSpanRefiningGenerator(RefiningGenerator):
         limit_lower=-np.inf,
         limit_upper=np.inf,
     ):
-        self.centre = centre
-        self.half_span = half_span
-        self.lower = centre - half_span
-        self.upper = centre + half_span
-        if self.lower < limit_lower and self.upper > limit_upper:
-            # If both extents of the span exceed the limits, change the span to fit
-            #   the larger of the two. This way we still don't move the scan centre.
-            self.half_span = max(abs(centre - limit_lower), abs(limit_upper - centre))
-            self.lower = centre - self.half_span
-            self.upper = centre + self.half_span
-        self.limit_lower = limit_lower
-        self.limit_upper = limit_upper
-        self.randomise_order = randomise_order
+        super().__init__(
+            *super().args_from_centre_span(
+                centre, half_span, randomise_order, limit_lower, limit_upper
+            )
+        )
+
+        self.centre, self.half_span = centre, half_span
+        self.limit_lower, self.limit_upper = limit_lower, limit_upper
 
 
 class ExpandingGenerator(ScanGenerator):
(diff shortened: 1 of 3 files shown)


ndscan/dashboard/scan_options.py
```diff
@@ -398,6 +402,7 @@ def write_to_params(self, params: dict) -> None:
                 "centre": self.box_centre.value() * self.scale,
                 "spacing": self.box_spacing.value() * self.scale,
                 "randomise_order": self.check_randomise.isChecked(),
+                "dtype": self.dtype,
```
dnadlinger: Can't we just query the parameter schema on the scan runner side?

test/test_experiment_scan_generator.py
```diff
@@ -158,48 +282,137 @@ def test_refininggenerator(self):
...
-        self.assertTrue(gen.has_level(3))
-        self.assertEqual(gen.points_for_level(3), [125.0, 175.0, 225.0, 275.0])
+        self.assertFalse(gen.has_level(3))
+        # We don't care whether 2 or 3 comes first or which ones are
+        # duplicated in level 2
+        points_levels_1_2 = set(gen.points_for_level(1) + gen.points_for_level(2))
+        self.assertIn(2, points_levels_1_2)
+        self.assertIn(3, points_levels_1_2)
+        all_points = points_levels_1_2.union(gen.points_for_level(0))
+        self.assertEqual(all_points, {1, 2, 3, 4})
```
dnadlinger: Surely, with `randomise_order=False`, the point order should be deterministic (and human-predictable)?

dnadlinger: (and non-duplicate!)

ayushagrawal2: The order is predictable, but from the perspective of a unit test it shouldn't matter whether 2 or 3 comes first as they're symmetric on either side of the midpoint. My thinking was that if someone reimplements this in a way that swaps the order, the unit test shouldn't fail as it's still a valid output

ayushagrawal2: Regarding duplicates, one would have to store all points already generated in the generator, which felt like it was against the point of it being a generator but it can be done.

Alternatively, it could be possible to predict which points have been taken by the previous levels if you know the spacing, but then you might run into issues with half-integers whose rounding would be sensitive to floating point errors. 

Do you have any suggestions

ndscan/experiment/scan_generator.py
```diff
@@ -35,17 +38,35 @@ def describe_limits(self, target: dict[str, Any]) -> None:
 class RefiningGenerator(ScanGenerator):
     """Generates progressively finer grid by halving distance between points each level."""
 
-    def __init__(self, lower, upper, randomise_order):
+    def __init__(self, lower, upper, randomise_order, dtype="float"):
```
dnadlinger: At the very least, these should be suitably defined enums rather than strings.

test/test_experiment_scan_generator.py
dnadlinger: > Regarding duplicates, one would have to store all points already generated in the generator, which felt like it was against the point of it being a generator but it can be done.

That's true, but also realistically isn't an issue – even just a simple set would be fine, as we typically store much more information per point anyway (in both the worker process, datasets, etc.), so the memory usage will be manageable.

I'd rather have a simple, predictable use experience (refining int scans visit each point at most once) than needing to document/explain weird behaviour contingent on the particular maths/rounding choices made in the implementation.

ndscan/dashboard/scan_options.py
ayushagrawal2: Where would we do that? Can't seem to find it

ndscan/experiment/scan_generator.py
```diff
@@ -232,6 +240,158 @@ def __init__(
...
+class IntCentreSpanRefiningGenerator(IntRefiningGenerator):
+    """Integer-valued analogue of :class:`CentreSpanRefiningGenerator`.
+
+    :param limit_lower: Optional lower limit (inclusive) to the range of generated
+            points.
+    :param limit_upper: See ``limit_lower``.
+    """
+
+    def __init__(self, *args, **kwargs):
+        CentreSpanRefiningGenerator.__init__(self, *args, **kwargs)
```
dnadlinger: Wow, this works?! I mean, logically, there isn't a reason why `self` in `CentreSpanRefiningGenerator.__init__(self, …)` would need to be a subtype of `CentreSpanRefiningGenerator`, because `CentreSpanRefiningGenerator.__init__` is just a funny name for a free function as far as Python 3 is concerned, but this almost seems a bit too clever. If I'm refactoring CentreSpanRefiningGenerator, I don't expect to have to look at a class that's not even a subclass, etc – i.e. this adds tight coupling between the implementation details of two classes where there conventionally is expected to be none.

Do you think it would be cleaner to have a free function that transforms the parameters e.g. from the centre/span refining spec to the upper/lower one (and same for the non-refining variant) and then just forward to the base class constructor?

ayushagrawal2: Fair point. A free function is one option. Alternatively, how about a static method in `RefiningGenerator`?
That way the function is nicely bundled with `RefiningGenerator`

```
class RefiningGenerator:
    @staticmethod
    def from_centre_span(instance, ...):
        instance.start = ...

class CentreSpanRefiningGenerator(RefiningGenerator):
    def __init__(...):
        RefiningGenerator.from_centre_span(self, ...)

class IntRefiningGenerator(RefiningGenerator):
    ...

class IntCentreSpanRefiningGenerator(IntRefiningGenerator):
    def __init__(...):
        IntRefiningGenerator.from_centre_span(self, ...)
```

ayushagrawal2: Or have that function return the transformed arguments so the deriving class explicitly calls `super().__init__(...)`

ayushagrawal2: does this look better now?

dnadlinger:
Thanks! I've moved the select_generator_class implementation to `entry_point`, as `scan_generator` can then be completely oblivious of the parameter type strings, etc. There was also a (presumably language server auto-inserted) `from venv import logger` rather than `logger = logging.getLogger(__name__)`.

ayushagrawal2:
Thanks!

dnadlinger:
Whoops, I must not have force-pushed to the right target or something like that; actually landed the logger fix in https://github.com/OxfordIonTrapGroup/ndscan/pull/495.

PR #496, opened 2026-05-12
Error handling for source_id label if underlying item is deleted

ayushagrawal2:
The source_id text item is sometimes updated after the parent is destroyed, spamming the console with errors.

image

### DIFF
--- a/ndscan/plots/plot_widgets.py
+++ b/ndscan/plots/plot_widgets.py
@@ -495,25 +495,32 @@ def add_source_id_label(
     """Add a translucent TextItem pinned to the bottom left of the view box displaying
     the context source id string.
     """
-    text_item = pyqtgraph.TextItem(
+
+    # Use custom slots instead of function pointers to prevent memory leak
+    # if the underlying C object is destroyed
+    class SourceIdTextItem(pyqtgraph.TextItem):
+        @QtCore.pyqtSlot(str)
+        def setText(self, text):
+            return super().setText(" " + text + " ")
+
+        @QtCore.pyqtSlot(object)
+        def setPosFromViewRange(self, viewBox):
+            (xmin, _), (ymin, _) = viewBox.viewRange()
+            return self.setPos(xmin, ymin)
+
+    text_item = SourceIdTextItem(
         text="", anchor=(0, 1), color=(255, 255, 255), fill=(0, 0, 0)
     )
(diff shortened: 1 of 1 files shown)


dnadlinger:
Thanks for looking into this!

The fix looks like it's just papering over a memory leak, though? In other words, this still leaves a memory leak in place, just silences the message that (usefully) points out that the object doesn't exist anymore. It seems the proper fix would be to remove the connection once the text item doesn't exist anymore (as the plot was recreated, presumably). This could be done by manually calling `disconnect` in the appropriate place, but probably more easily by connecting not to a Python "function pointer" but to a proper slot of a QObject that has e.g. `text_item` as a parent.

ayushagrawal2:
```
text_item.setText(" " + context.get_source_id() + " ")
```
Do we need the spaces before and after? We could redefine "setText" in a subclass of TextItem with spaces (even though that feels like a stupidly trivial subclass to make). Or we could add a padding in update_text_pos by mapping coordinates to and from local pixel coordinates, but that might suffer from the same memory leak issue

ayushagrawal2:
```python
def add_source_id_label(...) -> pyqtgraph.TextItem:
    class SourceIdTextItem(pyqtgraph.TextItem):
        def __init__(*args, padding_px = (5, 0), **kwargs):
            super().__init__(*args, **kwargs)
            self.padding_px = Point(padding_px)
    
        @QtCore.pyqtSlot(str)
        def setText(self, text):
            super().setText(" " + text + " ")
            
        @QtCore.pyqtSlot(object)
        def setPosFromViewRange(self, range):
            (xmin, _), (ymin, _) = range
            view_box = self.parent()
            view_box_corner_px = view_box.mapFromView(Point(xmin, ymin))
            padded_pos = view_box.mapToView(view_box_corner_px + self.padding_px)
            self.setPos(padded_pos)

    text_item = SourceIdTextItem(
        text="", anchor=(0, 1), color=(255, 255, 255), fill=(0, 0, 0)
    )
    ...

    context.source_id_changed.connect(text_item.setText)
    text_item.setText(context.get_source_id())
    view_box.sigRangeChanged.connect(text_item.setPosFromViewRange)
```

Something like this, where we can choose to keep either the spaces or the padding.

dnadlinger:
Thanks!

PR #507, opened 2026-08-07
Fix phantom cursor labels at negative coordinates for categorical scans

### LINKED ISSUE #506: plots: Repeated copy of enum cursor labels at negative "physical" axis coordinates
https://github.com/OxfordIonTrapGroup/ndscan/pull/503 made the cursors for categorical axes much more useful by resolving the floating point coordinates used for plotting back to categorical names.

However, currently a second phantom copy of the enum labels is displayed when hovering over the empty space towards more negative "physical" (pyqtgraph) axis coordinates values on an enum axis.

This is likely due to negative coordinates not being filtered out before casting to int and looking up the elements, so being interpreted as indexing from the end of the array.

@ayushagrawal2, could you have a quick look?

ayushagrawal2:
Fixes #506.

I've taken the liberty to cleanup the code a bit and make the cursor label consistent with the axis ticks (taking `.values()` of the `"members"` dict instead of `.keys()`. 

Also, since categorical data is only well-defined at the points where there is actually a tick, I've also added limits to the `view_box`es enclosing categorical scans preventing the user from scrolling out of range and ensuring there is at least always one valid category visible on the axis. Feel free to revert if you feel that's too restrictive for the user on the ux.

### DIFF
--- a/ndscan/plots/image_2d.py
+++ b/ndscan/plots/image_2d.py
@@ -25,6 +25,7 @@ from .utils import (
     find_neighbour_index,
     format_param_identity,
     get_axis_scaling_info,
+    get_param_categories,
     parse_coord_param_value,
     setup_axis_item,
     slice_data_along_axis,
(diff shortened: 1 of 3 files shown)


dnadlinger:
Thanks!

The view box limits don't appear to work for categorical x axes on a 2D scan. Looking at just the code here (not the pyqtgraph implementation), probably the two `setLimits` calls need to be merged into one?

ayushagrawal2:
Thanks for pointing that out! Haven't had a chance to test this yet, but hopefully that fixes it

Issue #29, opened 2019-02-09
plots: Slider to replay points as they come in


dnadlinger:
Possibly just overly cute rather than useful: Add a silder to the applet to replay the points in the order they came in, or two sliders on a timeline to filter chosen points by acquisition time.

PR #508, opened 2026-08-08
dashboard: Add time slider to scroll back through plot history

ayushagrawal2:
Adds a slider at the bottom of the widget to rollback through the point acquisition history in the order in which they came in -- useful e.g. for debugging if something in the experiment drifts.

The annotations are always the latest available from all points, and does not roll back, at least as currently implemented.

Open to suggestions for improving the UX / colour schemes etc. 
Should we e.g. hide the slider until explicitly enabled from a context menu?

This was originally motivated by MM saying she wanted to add progress bars for scans with deterministic number of points. I figured this could be cleanly done by a thin line going along the bottom of the screen, and would also be good to double as a time slider -- I've implemented only the latter here. 

Demo:

https://github.com/user-attachments/assets/b7c20e26-f652-4e6b-b92a-4a5ba36d5065

### DIFF
--- a/ndscan/plots/image_2d.py
+++ b/ndscan/plots/image_2d.py
@@ -11,10 +11,15 @@ from .._qt import QtCore, QtGui
 from . import colormaps
 from .cursor import CrosshairAxisLabel, CrosshairLabel, LabeledCrosshairCursor
 from .model import ScanModel
+from .model.history import HistoryFromScanModel
 from .model.select_point import SelectPointFromScanModel
 from .model.slice import create_slice_roots
 from .model.subscan import create_subscan_roots
-from .plot_widgets import SliceableMenuPanesWidget, add_source_id_label
+from .plot_widgets import (
+    SliceableMenuPanesWidget,
+    add_source_id_label,
+    add_time_slider,
+)
 from .utils import (
     CONTRASTING_COLOR_TO_HIGHLIGHT,
     HIGHLIGHT_PEN,
@@ -282,6 +287,16 @@ class _ImagePlot:
                 self.averages_by_coords[coords][0] if averaging_enabled else z
             )
 
+        # Remove any old points that have now been deleted
+        if num_to_show < self.num_shown:
+            all_x_inds = _coords_to_indices(x_data[:num_to_show], self.x_range)
+            all_y_inds = _coords_to_indices(y_data[:num_to_show], self.y_range)
+
+            removal_mask = np.full(self.image_data.shape, True)
+            removal_mask[all_x_inds, all_y_inds] = False
+
+            self.image_data[removal_mask] = np.nan
+
         cmap = colormaps.plasma
         channel = self.channels[self.active_channel_name]
         display_hints = channel.get("display_hints", {})
@@ -312,13 +327,17 @@ class Image2DPlotWidget(SliceableMenuPanesWidget):
     def __init__(self, model: ScanModel):
         super().__init__()
 
-        self.model = model
+        self.base_model = model
+        self.model = HistoryFromScanModel(model)
 
         self.model.channel_schemata_changed.connect(self._initialise_series)
         self.model.points_appended.connect(lambda p: self._update_points(p, False))
         self.model.points_rewritten.connect(lambda p: self._update_points(p, True))
 
         self.selected_point_model = SelectPointFromScanModel(self.model)
+        self.selected_point_model.source_index_changed.connect(
+            self._set_highlighted_index
+        )
 
         self.data_names = []
 
@@ -329,6 +348,7 @@ class Image2DPlotWidget(SliceableMenuPanesWidget):
         self.plot: _ImagePlot | None = None
         self.crosshair = None
         self._highlighted_xy = (None, None)
+        self.time_slider = None
 
         self.found_duplicate_coords = False
         self.unique_coords = set[tuple[float, float]]()
@@ -420,6 +440,9 @@ class Image2DPlotWidget(SliceableMenuPanesWidget):
 
         self.subscan_roots = create_subscan_roots(self.selected_point_model)
         self.slice_roots = create_slice_roots(self.model, self.selected_point_model)
+
+        if self.time_slider is None and self.base_model.has_independent_history:
+            self.add_time_slider()
 
         self.ready.emit()
 
@@ -528,8 +551,7 @@ class Image2DPlotWidget(SliceableMenuPanesWidget):
         y = self.plot.y_range[0] + y_idx * self.plot.y_range[2]
 
         source_idx = self._xy_to_source_index(x, y)
-        if source_idx is not None:
-            self._highlight_point_at_index(source_idx)
+        self.selected_point_model.set_source_index(source_idx)
 
     def keyPressEvent(self, event):
         """Handle arrow key presses to move the highlighted point."""
@@ -549,12 +571,11 @@ class Image2DPlotWidget(SliceableMenuPanesWidget):
         step = -1 if is_left or is_down else 1
         neighbour_idx = self._get_highlighted_neighbour_index(axis, step)
         if neighbour_idx is not None:
-            self._highlight_point_at_index(neighbour_idx)
+            self.selected_point_model.set_source_index(neighbour_idx)
         event.accept()
 
-    def _highlight_point_at_index(self, source_idx: int | None):
+    def _set_highlighted_index(self, source_idx: int | None):
         """Highlight the point at the given index of the source data."""
-        self.selected_point_model.set_source_index(source_idx)
 
         if source_idx is None:
             self._highlighted_xy = (None, None)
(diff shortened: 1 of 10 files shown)


dnadlinger:
Presumably, this should work on both 1D and 2D scans?

dnadlinger:
This is great to see, by the way (as is any movement on the plot UI front) – as evidenced by https://github.com/OxfordIonTrapGroup/ndscan/issues/29, this has been on my mind for a long time.

dnadlinger:
How does this interact with annotations/fitting? I'm not quite sure I'm able to tell at a glance from the video.

ayushagrawal2:
> Presumably, this should work on both 1D and 2D scans?

Yeah it should, I haven't implemented it for 2D scans yet but that should be quite quick to add

> How does this interact with annotations/fitting?

Currently it just changes the points shown -- the annotations/fits are taken independently of the slider from the full model with all points. 
I haven't set it up to save the annotations at each point. It shouldn't be too hard but I wasn't sure what the more useful behaviour would be

dnadlinger:
> Currently it just changes the points shown -- the annotations/fits are taken independently of the slider from the full model with all points.

That probably is the more useful behaviour anyway: Keep all the decorations stationary to provide a fixed (or slowly evolving) visual reference, and just hide/show the points.

ayushagrawal2:
This is almost ready now: here is the behaviour for a 2d scan with `ndscan_show`


https://github.com/user-attachments/assets/b4d3af63-09fe-48a3-a96a-80bb8b9b5cea


I've set it so the selected point is deselected and any slice panes are closed if the underlying point disappears. The minor problem with that, as seen in the second half of the video, is that this also expands the plot while the cursor is stationary, causing the slider to abruptly jump in value

dnadlinger:
On the slice panes/subscans, I think we should be able to keep them open, and just have "No data." displayed instead. This seems like the cleaner user experience when just quickly scrubbing back to check.

dnadlinger:
This is still marked as "Draft" – ready for review?

ayushagrawal2:
Think this is ready!

ndscan/plots/image_2d.py
```diff
@@ -654,3 +670,47 @@ def _set_view_box_limits(self, view_box):
             }
 
         view_box.setLimits(**kwargs)
+
+    def _create_time_slider(self):
+        # Avoid creating duplicates on series reinitialisation
+        if hasattr(self, "time_slider"):
+            return
```
dnadlinger: `ndscan` generally avoids `hasattr` in favour of more straight-line code (set to `None` in constructor, perhaps with a type annotation these days; check for `is None`).

ndscan/plots/image_2d.py
```diff
@@ -654,3 +670,47 @@ def _set_view_box_limits(self, view_box):
...
+
+        # Do not add slider to panes created by slices
+        if isinstance(self._base_model, SliceForScanModel):
+            self._has_time_slider = False
+            return
+
```
dnadlinger, line 686: Explicitly checking for object typing isn't very clean in the context of an OOP design such as the `model` interface here.

If we actually want to disallow this (acquisition order might be interesting for slices as well?), I'd prefer to do it on the plot widget side (e.g. argument passed down when creating it).

ndscan/plots/plot_widgets.py
```diff
@@ -488,6 +488,13 @@ def close_slice_plot(self, name):
         self.slice_plots[name].close()
         self.setFocus()
 
+    def close_all_slice_plots(self):
```
dnadlinger: Seems this is no longer needed.

ndscan/plots/xy_1d.py
```diff
@@ -634,3 +640,58 @@ def _set_view_box_limits(self, view_box):
...
+
+    def _create_time_slider(self):
+        # Avoid creating duplicates on series reinitialisation
+        if hasattr(self, "time_slider"):
+            return
+
+        # Do not add slider to panes created by slices
+        if isinstance(self._base_model, SliceForScanModel):
+            self._has_time_slider = False
+            return
```
dnadlinger: Same as for `image_2d`.

ndscan/plots/xy_1d.py
```diff
@@ -413,6 +418,7 @@ def _initialise_series(self, channels):
         self.ready.emit()
 
     def _update_points(self, points):
+        # print(points)
```
dnadlinger: Remove.

ndscan/plots/xy_1d.py
```diff
@@ -634,3 +640,58 @@ def _set_view_box_limits(self, view_box):
             x_min = -0.5
             x_max = len(param_categories) - 0.5
             view_box.setLimits(xMin=x_min, xMax=x_max, minXRange=1.05)
+
+    def _create_time_slider(self):
```
dnadlinger: What's the overlap to the `image_2d` case overall? If it's literally copy/paste, would it make sense to abstract out the setup code?

ndscan/plots/image_2d.py
ayushagrawal2: > acquisition order might be interesting for slices as well?

I think it's more sensible to have the slice state be tied to that of the base model, as opposed to having the two at different points in the timeline. The slice acquisition order is also implicitly accessible this way. If we want to decouple the two, I would rather hide that be opt-in via a context menu?

 > e.g. argument passed down when creating it
 
 In the RootWidget perhaps? Alternatively, we can have a class attribute `has_independent_history` in all `Model`s, which we set to false for the slice models

ndscan/plots/xy_1d.py
ayushagrawal2: It is just a copy/paste, so agreed, it would. Maybe a function that just creates the widget, and we let the plot widgets handle their placement, focus etc?

```python
def create_time_slider(
    history_model: HistoryFromScanModel,
    base_model: ScanModel,
):
    container = TimeSliderContainer()
    time_slider = container.slider
    time_slider.setSizePolicy(
        QtWidgets.QSizePolicy.Policy.Expanding,
        QtWidgets.QSizePolicy.Policy.Fixed,
    )

    time_slider.update_points(history_model.get_point_data())
    time_slider.state_changed.connect(history_model._update_state)

    base_model.points_appended.connect(time_slider.update_points)
    base_model.points_rewritten.connect(time_slider.update_points)

    return time_slider, container
```

ndscan/plots/plot_widgets.py
```diff
@@ -12,6 +12,9 @@
 import pyqtgraph
 import pyqtgraph.exporters
 
+from ndscan.plots.model.history import HistoryFromScanModel
+from ndscan.plots.time_slider import TimeSlider, TimeSliderContainer
```
dnadlinger: We generally use relative imports.

ndscan/plots/model/history.py
```diff
@@ -0,0 +1,135 @@
...
+    def get_channel_schemata(self) -> dict[str, Any]:
+        return self._channel_schemata
+
+    def get_point_data(self) -> dict[str, Any]:
+        return self._sliced_data
+
+    def quit(self) -> None:
+        self.parent.points_appended.disconnect(self._append_data)
+        self.parent.points_rewritten.disconnect(self._rewrite_data)
+        self.parent.annotations_changed.connect(self._update_annotations)
```
dnadlinger: Presumably, this should be `disconnect`.

ndscan/plots/model/history.py
```diff
@@ -0,0 +1,135 @@
...
+            self.points_rewritten.emit(self._sliced_data)
+        else:
+            self.points_appended.emit(self._sliced_data)
+
+    def _rewrite_data(self, *args) -> None:
+        point_data = self.parent.get_point_data()
+        num_points = len(next(iter(point_data.values()), []))
+
+        if self._state >= num_points:
+            self.set_state(-1)
```
dnadlinger: `update_state`?

ndscan/plots/model/history.py
```diff
@@ -0,0 +1,135 @@
...
+
+        self.update_state(self._state)
+
+    def _append_data(self, *args) -> None:
+        point_data = self.parent.get_point_data()
+        num_points = len(next(iter(point_data.values()), []))
+
+        if self._state >= num_points:
+            self.set_state(-1)
```
dnadlinger: `update_state`?

ndscan/plots/plot_widgets.py
```diff
@@ -489,6 +492,36 @@ def close_slice_plot(self, name):
         self.setFocus()
 
 
+def add_time_slider(
+    layout: QtWidgets.QGraphicsLayout,
+    scene: QtWidgets.QGraphicsScene,
+    model: HistoryFromScanModel,
+) -> TimeSlider:
+    """Add a slider at"""
```
dnadlinger: at where?

ndscan/plots/plot_widgets.py
ayushagrawal2: Apologies, auto-filled!

dnadlinger:
Great, thanks! Merged after coalescing into two atomic commits (models reset fit + the slider change), and renaming `state` to `cutoff` to avoid the former pretty vacuous term.

PR #522, opened 2026-09-02
Fix name collisions when saving similar OnlineFit analyses with different identifiers

ayushagrawal2:
The results channels for saving the OnlineFit params do not currently read the 'analysis_identifier', leading to name collisions when multiple fits of the same type are defined. Fixes this


dnadlinger:
Could you please add a unit test as a regression test for this?

